PR #12, opened 2022-01-28
Jcs/accounts table

jSylvestre:


Hippo.Core/Data/DbInitializer.cs
```diff
@@ -24,22 +24,56 @@ public async Task Initialize(bool recreateDb)
...
+            if (!(await _dbContext.Accounts.AnyAsync()))
+            {
+                var ownerId = (await _dbContext.Users.FirstAsync(a => a.Iam == "1000029584")).Id;
+                var account = new Account()
+                {
+                    CanSponsor = true,
+                    OwnerId = ownerId,
+                };
+                await _dbContext.Accounts.AddAsync(account);
+                await _dbContext.SaveChangesAsync();
```
srkirkland, line 54: Not that it matters but you don't need to save changes here, you can do it all at the end.  They way you would do it would be to set Sponsor to `account` instead of setting SponsorId via the re-query.   This is just dbInit code so I don't really care about this one but if it was anywhere else that's what i'd suggest

Hippo.Core/Domain/User.cs
```diff
@@ -32,5 +32,21 @@ public class User
 
         [StringLength(20)]
         public string Kerberos { get; set; }
+
+        [JsonIgnore]
+        public List<Account> Accounts { get; set; }
+
+        internal static void OnModelCreating(ModelBuilder modelBuilder)
+        {
+            modelBuilder.Entity<User>().HasIndex(a => a.Kerberos);
```
srkirkland, line 41: just mainly curiosity, but what's the point of indexing kerb?  Are we planning to search on it frequently or anywhere?  AFAIK we kinda just ignore it these days.

srkirkland (approved):

Hippo.Core/Domain/User.cs
jSylvestre: Harvest had it, just based it off that. I'll remove it and do another migration.

PR #16, opened 2022-01-31
Shhhh quiet time

jSylvestre:


Hippo.Web/Controllers/TestController.cs
```diff
@@ -42,5 +47,23 @@ public async Task<IActionResult> TestBody()
...
             return Content(results);
         }
+
+        public IActionResult TestSsh()
+        {
+            //read file into base64 string
+            //var file = System.IO.File.ReadAllBytes("D:\\Work\\GitProjects\\hippo\\Hippo.Web\\remote-api-auth.pk");
+            //var base64 = System.Convert.ToBase64String(file);
+
+            //return null;
```
srkirkland, line 57: don't need this anymore

Hippo.Core/Services/SshService.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        private readonly SshSettings _sshSettings;
+        private readonly PrivateKeyFile? _pkFile;
+
+        public SshService(IOptions<SshSettings> sshSettings)
+        {
+            _sshSettings = sshSettings.Value;
+            var rsa = Convert.FromBase64String(_sshSettings.Key);
+            var stream = new MemoryStream(rsa);
+            _pkFile = new PrivateKeyFile(stream);
```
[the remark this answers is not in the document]
jSylvestre: @srkirkland wrapping it seems to have worked fine.

Hippo.Core/Services/SshService.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    public interface ISshService
+    {
+        IEnumerable<string> Test();
+    }
+
+    public class SshService : ISshService
+    {
+        private readonly SshSettings _sshSettings;
+        private readonly PrivateKeyFile? _pkFile;
```
srkirkland, line 20: why is this nullable?  doesn't seem like it should be optional

srkirkland (changes requested):

PR #18, opened 2022-02-01
IdentityService

jSylvestre:


Hippo.Core/Hippo.Core.csproj
```diff
@@ -20,7 +20,6 @@
 	<ItemGroup>
 		<Folder Include="Migrations\SqlServer\" />
 		<Folder Include="Migrations\Sqlite\" />
-		<Folder Include="Utilities\" />
```
jSylvestre, line 23: Don't know why this was removed.... Maybe because it was an empty folder before?

srkirkland: don't know but you shouldn't need it

srkirkland (approved):
i can't quite tell what changed in startup.cs, but if it was just the auth stuff it looks good

jSylvestre:
> i can't quite tell what changed in startup.cs, but if it was just the auth stuff it looks good

@srkirkland 
Yes, just the auth and usings. If you hide whitespace, it is much easier to see:
image

PR #20, opened 2022-02-02
Show the username on the nav page

jSylvestre:


jSylvestre:
Pages need to be prettier, and not sure if I need all of that.
Copied and modified from harvest code

Hippo.Web/Views/_ViewImports.cshtml
```diff
@@ -1,2 +1,3 @@
 @using Hippo.Core.Domain
+@using Harvest.Web.Extensions
```
srkirkland, line 2: not harvest ;0

srkirkland (changes requested):
a few namespaces need to be changed

PR #28, opened 2022-02-03
example of uploading file to hpc

srkirkland:
also updated service code to clean up usings

### DIFF
--- a/Hippo.Core/Services/SshService.cs
+++ b/Hippo.Core/Services/SshService.cs
@@ -12,6 +12,7 @@ namespace Hippo.Core.Services
     public interface ISshService
     {
         IEnumerable<string> Test();
+        void PlaceFile(string contents, string path);
     }
 
     public class SshService : ISshService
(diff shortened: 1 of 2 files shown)


Hippo.Core/Services/SshService.cs
```diff
@@ -22,22 +23,40 @@ public class SshService : ISshService
...
         {
-            using (var client = new SshClient(_sshSettings.Url, _sshSettings.Name, _pkFile))
-            {
-                client.Connect();
-                var result = client.RunCommand("ls -l");
-                client.Disconnect();
-                return result.Result.Split('\n');
-            }
+            using var client = GetSshClient();
+            var result = client.RunCommand("ls -l");
```
jSylvestre, line 42: @srkirkland No Client disconnect / dispose?

jSylvestre: Nm, I missed that you had the using var...

jSylvestre (approved):

PR #30, opened 2022-02-04
download a file

jSylvestre:


Hippo.Core/Services/SshService.cs
```diff
@@ -58,5 +60,14 @@ private ScpClient GetScpClient()
             client.Connect();
             return client;
         }
+
+        public byte[] DownloadFile(string fileName)
+        {
+            using var client = GetScpClient();
+            using var stream = new MemoryStream();
+            client.Download(fileName, stream );
```
srkirkland, line 68: Interesting options here-- we can go with this for now but in general I would think you would want to return the stream and not byte[] for memory management reasons. Always good to not automatically load a entire file into memory if you can help it.

srkirkland (approved):

Hippo.Core/Services/SshService.cs
jSylvestre: If we have a using on the stream, would that work? Or would it get disposed before we could read it in the calling method?

PR #31, opened 2022-02-07
Cyd/design1

cydoval:
dark mode and some animation testing


Hippo.Web/ClientApp/src/components/Home.tsx
```diff
@@ -6,56 +6,36 @@ export class Home extends Component {
...
-          The <code>ClientApp</code> subdirectory is a standard React
-          application based on the <code>create-react-app</code> template. If
-          you open a command prompt in that directory, you can run{" "}
-          <code>npm</code> commands such as <code>npm test</code> or{" "}
-          <code>npm install</code>.
-        </p>
+      <div className="row justify-content-center">
+        <div className="col-md-6">
+          <h3>
+            Welcome, <span className="status-color">Calvin</span>
```
jSylvestre, line 12: I'm not sure how this will work with extending a component. Maybe @srkirkland has a suggestion?

srkirkland: We'll just change to `const Home = () => {}` and the use `useContext()`.  And we'll likely break this into other component anyway -- we aren't going to want any of this logic directly in home.  But it's a good start style/design wise i think and we can change those after the PR.

srkirkland (approved):

PR #32, opened 2022-02-07
Change to function, use user's name

jSylvestre:


Hippo.Web/ClientApp/src/components/Home.tsx
```diff
@@ -1,42 +1,41 @@
-import React, { Component } from "react";
+import { Component, useContext } from "react";
```
jSylvestre, line 1: Delete Component? Not being used as far as I can tell.

srkirkland (approved):

PR #36, opened 2022-02-09
Emulate

jSylvestre:


Hippo.Web/Controllers/SystemController.cs
```diff
@@ -0,0 +1,80 @@
...
+            _identityService = identityService;
+            _userService = userService;
+        }
+        public async Task<IActionResult> Emulate(string id)
+        {
+            var allowedUsers = new[] {"jsylvest", "postit", "cydoval", "sweber" };
+            var currentUser = await _userService.GetCurrentUser();
+            if(currentUser == null || !allowedUsers.Contains(currentUser.Kerberos))
+            {
+                throw new Exception("Not authorized");
```
srkirkland, line 31: just an FYI this is fine for emulation but for regular routes in this scenario you'd want to `return NotAuthorized()` instead of throwing an exception

srkirkland (approved):
Let's make sure to throw in logging for both methods once logging is added to the project

Hippo.Web/Controllers/SystemController.cs
jSylvestre: @srkirkland I didn't see a NotAuthorized() method, but there is UnAuthorized().

PR #39, opened 2022-02-10
Decision email

jSylvestre:
image

image

### DIFF
--- a/Hippo.Core/Services/EmailService.cs
+++ b/Hippo.Core/Services/EmailService.cs
@@ -104,7 +104,7 @@ namespace Hippo.Core.Services
         {
             if(account.Owner == null || account.Sponsor == null || account.Sponsor.Owner == null)
             {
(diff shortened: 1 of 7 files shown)


Hippo.Core/Services/EmailService.cs
```diff
@@ -34,9 +34,41 @@ public EmailService(AppDbContext dbContext, INotificationService notificationSer
             _emailSettings = emailSettings.Value;
         }
 
-        public Task<bool> AccountDecission(Account account, bool isApproved)
+        public async Task<bool> AccountDecission(Account account, bool isApproved)
         {
-            throw new NotImplementedException();
+            try
+            {
+                account = await CheckForMissingData(account);
```
srkirkland, line 41: I don't really like the name or maybe the functionality of this, I can't quite decide.  One option would be to have `EnsureAccountData(account)` and then it just updates in-place.  Another would be something like `GetCompleteAccount(account)` which could return the account like it does now.  Something about `CheckFor` returning an object you need to use to modify your param doesn't quite sit well.

srkirkland (changes requested):

Hippo.Core/Services/EmailService.cs
jSylvestre: Ha, I had something similar to ensure, then I changed it to what I had for harvest :)

PR #44, opened 2022-02-16
Log IAM issues

jSylvestre:


srkirkland (approved):
This is useful but it makes me think what happens if we get into those else conditions. It isn't a valid state for us to just log that no IAM is found. Perhaps redirect to some page that logs you out and says contact support because cas is having problems?  Not a unique to hippo problem of course.

PR #45, opened 2022-02-16
Jcs/is active change

jSylvestre:


sprucely (approved):
Makes sense that an account should be active when it's created.

srkirkland (approved):

srkirkland (approved):

PR #47, opened 2022-02-17
change to url navigation

srkirkland:
home page now just redirects to different pages depending on user account status.  those pages are loaded via react-router instead of via IF statements.  This also allows the app.tsx file to change className based on the url.  

We will probably need smarter functions for the className translations in the future and we can probably come up with better route or component names, but this does the trick for now


jSylvestre (approved):
🚢

PR #51, opened 2022-02-17
save actual ssh file to remote location

srkirkland:
test of sending the ssh key to farm on approval.  it does work.  still needs emailing and history

### DIFF
--- a/Hippo.Web/Controllers/AccountController.cs
+++ b/Hippo.Web/Controllers/AccountController.cs
@@ -12,11 +12,13 @@ public class AccountController : SuperController
 {
     private AppDbContext _dbContext;
     private IUserService _userService;
+    private ISshService _sshService;
 
-    public AccountController(AppDbContext dbContext, IUserService userService)
+    public AccountController(AppDbContext dbContext, IUserService userService, ISshService sshService)
     {
         _dbContext = dbContext;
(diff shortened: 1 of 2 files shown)


Hippo.Web/Controllers/AccountController.cs
```diff
@@ -49,15 +51,23 @@ public async Task<ActionResult> Approve(int id)
...
+            .SingleOrDefaultAsync(a => a.Id == id && a.Sponsor.OwnerId == currentUser.Id && a.Status == Account.Statuses.PendingApproval);
 
         if (account == null)
         {
             return NotFound();
         }
 
+        Console.WriteLine($"Approving account {account.Owner.Iam} with ssh key {account.SshKey}");
+
+        _sshService.PlaceFile(account.SshKey, $"/var/lib/remote-api/{account.Owner.Kerberos}.txt");
```
jSylvestre, line 64: @srkirkland What about using the IAM id?

jSylvestre (approved):

Hippo.Web/Controllers/AccountController.cs
srkirkland: they want kerbID.  BUT they want it "in" the file so the name will probably change anyway

PR #57, opened 2022-02-18
Show error, use callback

jSylvestre:


Hippo.Web/ClientApp/src/components/RequstForm.tsx
```diff
@@ -42,6 +42,10 @@ export const RequestForm = () => {
       const newAccount = await response.json();
       setContext((ctx) => ({ ...ctx, account: newAccount }));
       history.replace("/"); // could also push straight to pending, but home will redirect there immediately anyway
+    } else {
+      const errorText = await response.text();
```
srkirkland, line 46: I'm curious why would the error come back as text? Is this just like a full text dump of the dev exception page?  In prod it wouldn't be very useful. Unless we are sending back specific error codes or maybe im missing something?

jSylvestre:
Resolves #58

PR #63, opened 2022-02-22
Upload and rename

### LINKED ISSUE #53: Send file in temp directory and then rename/move file
Whenever creating a file via SSH, create it first either in a temp folder or as a "dot" file.  Then rename/move to the proper endpoint directory since the rename is an atomic operation

jSylvestre:
Resolve #53

### DIFF
--- a/Hippo.Web/Controllers/AccountController.cs
+++ b/Hippo.Web/Controllers/AccountController.cs
@@ -66,10 +66,11 @@ public class AccountController : SuperController
 
         Console.WriteLine($"Approving account {account.Owner.Iam} with ssh key {account.SshKey}");
 
+        var tempFileName = $"/var/lib/remote-api/.{account.Owner.Kerberos}.txt"; //Leading .
         var fileName = $"/var/lib/remote-api/{account.Owner.Kerberos}.txt";
 
-        _sshService.PlaceFile(account.SshKey, $".{fileName}"); //Leading .
-        _sshService.RenameFile($".{fileName}", $"{fileName}");
+        _sshService.PlaceFile(account.SshKey, tempFileName); 
+        _sshService.RenameFile(tempFileName, fileName);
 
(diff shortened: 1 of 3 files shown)


Hippo.Web/Controllers/AccountController.cs
```diff
@@ -66,7 +66,10 @@ public async Task<ActionResult> Approve(int id)
 
         Console.WriteLine($"Approving account {account.Owner.Iam} with ssh key {account.SshKey}");
 
-        _sshService.PlaceFile(account.SshKey, $"/var/lib/remote-api/{account.Owner.Kerberos}.txt");
+        var fileName = $"/var/lib/remote-api/{account.Owner.Kerberos}";
+
+        _sshService.PlaceFile(account.SshKey, $"{fileName}.dot");
+        _sshService.RenameFile($"{fileName}.dot", $"{fileName}.txt");
```
sprucely, line 72: I think this should be
```csharp
        _sshService.PlaceFile(account.SshKey, $".{fileName}");
        _sshService.RenameFile($".{fileName}", $"{fileName}.txt");
```

Hippo.Web/Controllers/TestController.cs
```diff
@@ -99,6 +99,28 @@ public IActionResult TestScd()
...
+        public IActionResult TestRename()
+        {
+            var testValue = _sshService.Test();
+            if(testValue.Contains("jcstest.txt"))
+            { 
+                _sshService.RenameFile("jcstest.txt", "jcstest.dot");
+            }
+            else
+            {
+                _sshService.RenameFile("jcstest.dot", "jcstest.txt");
```
sprucely, line 111: Same here. Prefix file with a `.` instead of a suffix of `.dot`

sprucely (changes requested):

PR #71, opened 2022-02-28
Jcs/admin account approval

jSylvestre:
image

### DIFF
--- a/Hippo.Web/ClientApp/src/Admin/AdminApproveAccounts.tsx
+++ b/Hippo.Web/ClientApp/src/Admin/AdminApproveAccounts.tsx
@@ -1,0 +1,110 @@
+import { useEffect, useState } from "react";
+
+import { Account } from "../types";
+
+import { authenticatedFetch } from "../util/api";
+
+export const AdminApproveAccounts = () => {
+  // get all accounts that need approval and list them
+  // allow user to approve or reject each account
+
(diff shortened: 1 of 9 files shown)


Hippo.Web/ClientApp/src/Admin/AdminApproveAccounts.tsx
```diff
@@ -0,0 +1,117 @@
...
+                  <td>{account.name}</td>
+                  <td>
+                    {account.sponsor?.name} ({account.sponsor?.owner?.email})
+                  </td>
+                  <td>{new Date(account.createdOn).toLocaleDateString()}</td>
+                  <td>
+                    <button
+                      disabled={
+                        accountApproving !== undefined &&
+                        accountRejecting !== undefined
```
sprucely, line 87: Wouldn't this be a little cleaner with `accountApproving || accountRejecting`? Or is it done this way because it's possible for an account id to be `0`?

sprucely (approved):

Hippo.Web/ClientApp/src/Admin/AdminApproveAccounts.tsx
jSylvestre: Scott originally just had 1 button and `disabled={accountSubmitting !== undefined}`
I added another button for rejecting and just kept that pattern.

Issue #61, opened 2022-02-18
Allow sponsor to reject a request


srkirkland:
Probably require reject comment/reason, send email with that to user, set account request to IsActive = false

jSylvestre:
Didn't do a comment with that merge.
Have to think about that a little.

PR #73, opened 2022-03-01
Jcs/reject reason

jSylvestre:


jSylvestre:
resolve #61

jSylvestre:
image

Hippo.Web/ClientApp/src/Shared/RejectRequest.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <textarea
+            className="form-control"
+            id="fieldName"
+            rows={3}
+            required
+            value={reason}
+            onChange={(e) => {
+              setReason(e.target.value);
+              setReturn(e.target.value);
+            }}
```
sprucely, line 37: Should we do something here to enforce the 1500 char limit? Or should we wait until we have yup validation in place?

sprucely (approved):

Hippo.Web/ClientApp/src/Shared/RejectRequest.tsx
jSylvestre: That was me just pulling a number out of my hat... Email would still get the full text

PR #76, opened 2022-03-03
When CAS doesn't return claims

jSylvestre:
Happening too often


jSylvestre:
I could move this into the userService method (with a little work) so it only calls IAM if the IAM id is missing or it can't find the user in the db

sprucely (approved):
I think this is fine. If you were to move this into a service, I think `IdentityService` would be more appropriate than UserService. Something like 
```
async Task EnsureValidClaims(identity: ClaimsIdentity) {...}
```

PR #102, opened 2022-03-11
added / updated, and show error message if bad request

jSylvestre:


Hippo.Web/ClientApp/src/components/Admin/Sponsors.tsx
```diff
@@ -53,38 +53,52 @@ export const Sponsors = () => {
...
+              : [newAccount];
+            message = "Sponsor Added";
+          }
+          //sort the list
+          setAccounts(
+            updatedAccounts.sort((a, b) => a.name.localeCompare(b.name))
+          );
+
+          setRequest((r) => ({ ...r, lookup: "" }));
+          setRequest((r) => ({ ...r, name: "" }));
```
srkirkland, line 86: don't do two in a row, do them both together (same reason as before).  `setRequest(r=>({ ...r, lookup: "", name: ""}));`

jSylvestre:
no merge

PR #104, opened 2022-03-14
buttons on table, animation

cydoval:


srkirkland (approved):

Hippo.Web/ClientApp/src/components/Account/AccountInfo.tsx
```diff
@@ -6,7 +6,7 @@ export const AccountInfo = () => {
 
   return (
     <div className="row justify-content-center">
-      <div className="col-md-6">
+      <div className="8">
```
jSylvestre, line 9: @cydoval just 8 ?

PR #106, opened 2022-03-17
Jcs/svg cleanup

jSylvestre:


jSylvestre:
Deleting PR to avoid conflicts

PR #118, opened 2022-03-18
rsa public key and pending page text

### LINKED ISSUE #110: Public, not private key

jSylvestre:
Closes #110


Hippo.Web/ClientApp/src/components/Account/PendingApproval.tsx
```diff
@@ -9,7 +9,8 @@ export const PendingApproval = () => {
       <div className="col-md-8">
         <p>
           Welcome {context.user.detail.firstName} your account is pending
-          approval. Please wait...
+          approval. Your sponsor has been emailed and your will be notified when
```
srkirkland, line 12: typo, should be "you* will be notified"

PR #119, opened 2022-03-18
Jcs/yml ssh

### LINKED ISSUE #107: YAML format for uploaded file on account approval
Need to convey 
1. account kerb
2. sponsor kerb
3. account PK

account kerb can be the name of the file, like it is now, so it's optional if we need to put it in the file.  we might as well though i think.

jSylvestre:
Closes #107

image

### DIFF
--- a/Hippo.Web/Controllers/AccountController.cs
+++ b/Hippo.Web/Controllers/AccountController.cs
@@ -164,7 +164,7 @@ public class AccountController : SuperController
             return BadRequest("Invalid SSH key");
         }
 
-        var sponsorAccount = await _dbContext.Accounts.Include(a => a.Owner).SingleAsync(a => a.OwnerId == model.SponsorId);
+        var sponsorAccount = await _dbContext.Accounts.Include(a => a.Owner).SingleAsync(a => a.Id == model.SponsorId);
 
 
         var sb = new StringBuilder();
--- a/Hippo.Web/Controllers/AccountController.cs
+++ b/Hippo.Web/Controllers/AccountController.cs
@@ -174,14 +174,14 @@ public class AccountController : SuperController
         sb.AppendLine($"    email: {sponsorAccount.Owner.Email}");
         sb.AppendLine($"    kerb: {sponsorAccount.Owner.Kerberos}");
         sb.AppendLine($"    iam: {sponsorAccount.Owner.Iam}");
-        sb.AppendLine($"    mothra: TODO"); //TODO
+        sb.AppendLine($"    mothra: {sponsorAccount.Owner.MothraId}"); 
         sb.AppendLine();
         sb.AppendLine("Account:");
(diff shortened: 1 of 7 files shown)


Hippo.Web/Controllers/AccountController.cs
```diff
@@ -158,17 +159,39 @@ public async Task<ActionResult> Create([FromBody] CreateModel model)
...
+        sb.AppendLine($"    iam: {sponsorAccount.Owner.Iam}");
+        sb.AppendLine($"    mothra: {sponsorAccount.Owner.MothraId}"); 
+        sb.AppendLine();
+        sb.AppendLine("Account:");
+        sb.AppendLine($"    name: {currentUser.Name}");
+        sb.AppendLine($"    email: {currentUser.Email}");
+        sb.AppendLine($"    kerb: {currentUser.Kerberos}");
+        sb.AppendLine($"    iam: {currentUser.Iam}");
+        sb.AppendLine($"    mothra: {currentUser.MothraId}"); 
+        sb.AppendLine($"    key: {model.SshKey}");
```
sprucely, line 185: It's hard to tell from the screenshot if there are line breaks in the `key` property or if it is just wrapping. If there are line breaks, then it's not valid yaml. If it is a single line, you'll want to enclose it in double-quotes to prevent any special characters from being interpreted as yaml string-formatting directives.

A double-quoted string is the only way to have a block-formatted yaml string while preventing the yaml parser from inserting whitespace. This...
```yaml
key : "BEGIN

  abc\
  123\
  @#$%

  END"
```
Will result in a string of `"BEGIN\nabc123@#$%\nEND"`

You can test yaml at https://yaml-online-parser.appspot.com/
[Here's an exhaustive explanation of yaml string formatting](https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines)

sprucely (changes requested):

Hippo.Web/Controllers/AccountController.cs
jSylvestre: The line breaks from my first example were because that was how I pasted it in.
Although, I suppose we could validate if they pasted them in...

srkirkland: As long as it passes that yml validation i'm good with it.  does it work w/ your example?  can't tell about the line breaks, and I don't think you need to quite multi-line in YAML but it's worth trying if the first version fails and you get stuck

srkirkland (commented):
Everything looks good as long as the yaml validates properly

srkirkland:
Maybe i'd suggest a YAML service or at least private method for the big yaml creation thing.  I knows it's only in one place now but that takes up a lot of the create method of the account controller with details that aren't really needed if you are skimming what create does

Issue #115, opened 2022-03-17
v2 - multi-cluster support 


srkirkland:
Add something to accounts maybe which have cluster attachment?  new clusters table?

srkirkland:
if so, our routes will need to be updated to point to new cluster, like "team" in peaks.  would need way to store SSH login info per-cluster.

jSylvestre:
Related to issue #15

jSylvestre:
- [ ] Create new db tables and fields
- [ ] Update URLs to contain cluster name
- [ ] Use cluster name to determine which account to load
- [ ] use account to determine what to show
- [ ] use cluster and account to determine server side permissions with the policy

PR #124, opened 2022-03-21
Db structure for Cluster

jSylvestre:
issue #115

### DIFF
--- a/Hippo.Core/Migrations/SqlServer/AppDbContextSqlServerModelSnapshot.cs
+++ b/Hippo.Core/Migrations/SqlServer/AppDbContextSqlServerModelSnapshot.cs
@@ -33,10 +33,19 @@ namespace Hippo.Core.Migrations.SqlServer
                     b.Property<bool>("CanSponsor")
                         .HasColumnType("bit");
 
+                    b.Property<int>("ClusterId")
+                        .HasColumnType("int");
+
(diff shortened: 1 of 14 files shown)


jSylvestre:
@srkirkland I assume that for auth to work with all of this the URL string will always have to contain the cluster.name?
Then we can extract that out like we do the projectId in harvest?

jSylvestre:
Think I did something wrong. I'm seeing a clusterId and a clusterId1 in the account...

srkirkland (approved):

Hippo.Core/Migrations/SqlServer/AppDbContextSqlServerModelSnapshot.cs
```diff
@@ -245,15 +240,11 @@ protected override void BuildModel(ModelBuilder modelBuilder)
             modelBuilder.Entity("Hippo.Core.Domain.Account", b =>
                 {
                     b.HasOne("Hippo.Core.Domain.Cluster", "Cluster")
-                        .WithMany()
+                        .WithMany("Accounts")
                         .HasForeignKey("ClusterId")
-                        .OnDelete(DeleteBehavior.Restrict)
+                        .OnDelete(DeleteBehavior.Cascade)
```
jSylvestre, line 245: Restrict instead?

PR #127, opened 2022-03-22
Multi-cluster support

srkirkland:
Handles structure of multi-cluster support, including all user activity and some example admin activity.

All API endpoints client/server have been updated to take cluster as url param.  Components have been updated to check this param and controllers expose it.  LINQ queries now use custom `InCluster()` filter for most queries.

A few admin functions still need to be updated (see TODO in adminController.cs).  Additionally admin permissions checked by auth policy need to be updated to handle cluster as well.

### DIFF
--- a/Hippo.Core/Models/AccountDetail.cs
+++ b/Hippo.Core/Models/AccountDetail.cs
@@ -7,6 +7,7 @@ namespace Hippo.Core.Models
         public int Id { get; set; }
         public string Name { get; set; }
         public string Status { get; set; }
+        public bool CanSponsor { get; set; }
         public string Owner { get; set; }
         public string Cluster { get; set; }
         public string Sponsor { get; set; }
--- a/Hippo.Core/Models/AccountDetail.cs
+++ b/Hippo.Core/Models/AccountDetail.cs
@@ -11,5 +11,6 @@ namespace Hippo.Core.Models
         public string Owner { get; set; }
         public string Cluster { get; set; }
         public string Sponsor { get; set; }
+        public bool IsAdmin { get; set; }
     }
 }
--- a/Hippo.Core/Models/AccountDetail.cs
+++ b/Hippo.Core/Models/AccountDetail.cs
@@ -1,0 +1,14 @@
+using Hippo.Core.Domain;
+
+namespace Hippo.Core.Models
+{
+    public class AccountDetail
+    {
+        public int Id { get; set; }
+        public string Name { get; set; }
+        public string Status { get; set; }
+        public string Owner { get; set; }
+        public string Cluster { get; set; }
+        public string Sponsor { get; set; }
+    }
+}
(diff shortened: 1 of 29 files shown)


Hippo.Core/Models/AccountDetail.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public class AccountDetail
+    {
+        public int Id { get; set; }
+        public string Name { get; set; }
+        public string Status { get; set; }
+        public bool CanSponsor { get; set; }
+        public string Owner { get; set; }
+        public string Cluster { get; set; }
+        public string Sponsor { get; set; }
```
jSylvestre, line 13: Need IsAdmin

Hippo.Web/ClientApp/src/components/Account/RequestForm.tsx
```diff
@@ -66,7 +72,7 @@ export const RequestForm = () => {
           <span className="status-color">{context.user.detail.firstName}</span>
         </h3>
         <p>
-          You don't seem to have an account on Farm yet. If you'd like access,
+          You don't seem to have an account on the {cluster} cluster yet. If you'd like access,
```
jSylvestre, line 75: Worthwhile getting the cluster description here?

Hippo.Web/ClientApp/src/components/Admin/AdminUsers.tsx
```diff
@@ -117,17 +119,17 @@ export const AdminUsers = () => {
...
+                  <td>{account.owner?.email}</td>
                   <td>
                     <button
                       disabled={notification.pending}
-                      onClick={() => handleRemove(user)}
+                      onClick={() => handleRemove(account)}
                       className="btn btn-primary"
                     >
-                      {adminRemoving === user.id ? "Removing..." : "Remove"}
+                      {adminRemoving === account.id ? "Removing..." : "Remove"}
```
jSylvestre, line 132: I think this still needs to be the user id? account.owner.id

jSylvestre (approved):

Hippo.Web/ClientApp/src/components/Account/RequestForm.tsx
srkirkland: i didn't think it was worth it.  but if we put it in the context then maybe..

Hippo.Web/ClientApp/src/components/Admin/AdminUsers.tsx
srkirkland: not sure but we probably want to move to account ids and away from owner useage

jSylvestre: @srkirkland This was my mistake. This needs to be account.id. But I had to change the post to use user.id.
Fixed in my pr

PR #130, opened 2022-03-23
Jcs/cluster f ixes20220323

jSylvestre:


Hippo.Web/Controllers/AdminController.cs
```diff
@@ -87,13 +87,13 @@ public async Task<IActionResult> Create(string id)
                 IsAdmin = true,
                 IsActive = true,
                 Status = Account.Statuses.Active,
-                ClusterId = (await _dbContext.Clusters.SingleAsync(c => c.Name == Cluster)).Id
+                Cluster = await _dbContext.Clusters.SingleAsync(c => c.Name == Cluster)
```
srkirkland, line 90: if this works i'm ok with it.  But I've seen most docs just try to pull out the ID which would be more efficient but also would save some object tracking and potential bad saves.  Perhaps best case it would look like 
```
ClusterId = await _dbContext.Clusters.Where(c => c.Name == Cluster).Select(c=>c.Id).SingleAsync();
```

srkirkland (approved):

Hippo.Web/Controllers/AdminController.cs
jSylvestre: I did it like this because the history is using cluster, not clusterid. 
Might be worth refactoring it to just use the id...

PR #136, opened 2022-03-28
some minor polish

srkirkland:


Hippo.Web/ClientApp/src/AppNav.tsx
```diff
@@ -14,7 +14,9 @@ export const AppNav = () => {
     <div>
       <div className="row appheader justify-content-center">
         <div className="col-md-8 hippo">
-          <HippoLogo />
+          <NavLink to="/">
```
jSylvestre, line 17: @srkirkland When we have more than 1 cluster, this should go to:
/clusters

jSylvestre (approved):

PR #137, opened 2022-03-28
Change account owner.

jSylvestre:
Untested....


srkirkland (approved):
Didn't try it but this seems to me like a good way to do the sponsor account change instead of updating all sub accounts.

Issue #145, opened 2022-04-15
Bulk load?


jSylvestre:
Page to bulk load accounts.

Maybe kerb-sponsorkerb,kerb-sponsorkerb,kerb-sponsorkerb,

Then they could go in an clean up any sponsor names?

jSylvestre:
No longer applicable with groups....

PR #146, opened 2022-04-15
Bulk load service

jSylvestre:
Mostly tested

### DIFF
--- a/Hippo.Web/Services/BulkLoadService.cs
+++ b/Hippo.Web/Services/BulkLoadService.cs
@@ -37,8 +37,8 @@ namespace Hippo.Web.Services
             foreach(var pair in pairs)
             {
                 var kerbIs = pair.Split("-");
-                kerbs.Add(kerbIs[0]);
-                kerbs.Add(kerbIs[1]);
+                kerbs.Add(kerbIs[0].Trim());
+                kerbs.Add(kerbIs[1].Trim());
             }
             kerbs = kerbs.Distinct().ToList();
             Log.Information($"Unique Kerbs: {kerbs.Count()}");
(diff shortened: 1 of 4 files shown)


jSylvestre:
Maybe load all the sponsor accounts first....

jSylvestre:
Issue #145

Hippo.Web/Services/BulkLoadService.cs
```diff
@@ -0,0 +1,148 @@
...
+                    Log.Error($"Kerb not found: {kerb}");
+                    continue;
+                }
+                if(!await _dbContext.Users.AnyAsync(a => a.Iam == user.Iam))
+                {
+                    Log.Information($"Adding user: {kerb}");
+                    await _dbContext.Users.AddAsync(user);
+                }
+                kerbsInDb.Add(kerb);
+            }
```
sprucely, line 63: Not sure if multiple calls to identity service can be avoided, but multiple calls to db certainly can without getting fancy with udf/temp table stuff if you load all the kerbs into memory (swap Iam for Id if you want to search by that) 
```csharp
var currentKerbs = _dbContext.Users.Select(x => x.Id).ToHashSet();
var kerbsToBeAdded = kerbs.Where(id => !currentKerbs .Contains(id));
```
I wouldn't expect the number of user records to ever get high enough to cause a memory issue.

Hippo.Web/Services/BulkLoadService.cs
```diff
@@ -0,0 +1,148 @@
...
+        public async Task<int> Load(Cluster cluster, string data)
+        {
+            var count = 0;
+            var pairs = data.Split(",");
+
+            Log.Information($"Starting bulk load for {pairs.Count()} accounts");
+
+            //Create missing users
+            var kerbs = new List<string>();
+            var kerbsInDb = new List<string>();
```
sprucely, line 36: This should probably be a `HashSet<string>` if you're going to be calling `.Contains(...)` on it in a loop.

sprucely (approved):
Haven't tested anything, but it looks okay to me. Left a couple comments about things that can be done to improve performance.

jSylvestre:
Closing without merging as this wouldn't work with the new way accounts will be done.

PR #147, opened 2022-06-15
Add support for cluster-specific ssh keys

sprucely:


Hippo.Web/Controllers/ClusterAdminController.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        [HttpPost]
+        public async Task<IActionResult> Delete(int id)
+        {
+            var cluster = await _dbContext.Clusters.AsNoTracking().SingleOrDefaultAsync(c => c.Id == id);
+            if (cluster == null)
+            {
+                return NotFound();
+            }
+            _dbContext.Remove(cluster);
```
jSylvestre, line 129: @sprucely Probably don't want to do this if it has any accounts.

jSylvestre: And probably an inactive flag...

Hippo.Core/Domain/Cluster.cs
```diff
@@ -20,12 +20,20 @@ public class Cluster
         [Required]
         [MaxLength(250)]
         public string Description { get; set; } = String.Empty;
+        [MaxLength(250)]
+        public string SshName { get; set; } = String.Empty;
+        [MaxLength(40)]
+        public string SshKeyId { get; set; } = String.Empty;
+        [MaxLength(250)]
+        public string SshUrl { get; set; } = String.Empty;
+        public bool IsInactive { get; set; }
```
jSylvestre, line 29: Default to true?

jSylvestre: Oh, you did IsInactive.
We usually do IsActive...

Hippo.Web/ClientApp/src/components/ClusterAdmin/Clusters.tsx
```diff
@@ -0,0 +1,324 @@
...
+    {
+      title: "Remove Cluster",
+      message: "Are you sure you want to remove this cluster?",
+    },
+    []
+  );
+
+  useEffect(() => {
+    const fetchClusters = async () => {
+      const response = await authenticatedFetch(`/api/_/clusteradmin/clusters`);
```
srkirkland, line 152: what's this strange underscore url thing?

Hippo.Core/Domain/Cluster.cs
sprucely: Why would you create a cluster unless you intend it to be active?

sprucely: Also, if you create an inactive cluster, it will be automatically filtered out of results.

jSylvestre: I didn't read your var name correctly. All our other apps use IsActive name

Hippo.Web/Controllers/ClusterAdminController.cs
```diff
@@ -0,0 +1,134 @@
...
+        [HttpGet]
+        public async Task<IActionResult> Clusters()
+        {
+            var currentSecretKeyIds = new HashSet<string>(await _secretsService.GetSecretNames());
+
+            var clusterModels = (await _dbContext.Clusters
+                .AsNoTracking()
+                .OrderBy(c => c.Name)
+                .ToArrayAsync())
+                .Select(c => new ClusterModel(c, string.Empty))
```
srkirkland, line 39: this is interesting -- i'm sure there is something that would fail if you didn't, but it seems unfortunate you have to ToArray() which executes the query, and then do a .Select() which could only need a subset of that data.

Hippo.Web/Controllers/ClusterAdminController.cs
```diff
@@ -0,0 +1,134 @@
...
+                .ToArray();
+
+            foreach (var model in clusterModels)
+            {
+                if (!string.IsNullOrEmpty(model.Cluster.SshKeyId) && currentSecretKeyIds.Contains(model.Cluster.SshKeyId))
+                {
+                    model.SshKey = await _secretsService.GetSecret(model.Cluster.SshKeyId);
```
srkirkland, line 46: why do we need to get the actual key for each cluster?  shouldn't we only ever get the key at the moment we are going to push a file via ssh?

Hippo.Web/Controllers/ClusterAdminController.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        [HttpGet]
+        public async Task<IActionResult> Cluster(int id)
+        {
+            var cluster = await _dbContext.Clusters.AsNoTracking().SingleOrDefaultAsync(c => c.Id == id);
+            if (cluster == null)
+            {
+                return NotFound();
+            }
+            var sshKey = !string.IsNullOrEmpty(cluster.SshKeyId)
```
srkirkland, line 61: also feels like maybe we don't want to get the ssh key here either

Hippo.Web/Startup.cs
```diff
@@ -186,9 +186,9 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<INotificationService, NotificationService>();
             services.AddScoped<ISshService, SshService>();
             services.AddScoped<IUserService, UserService>();
-            services.AddSingleton<IHttpContextAccessor, NullHttpContextAccessor>();
+            services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```
srkirkland, line 189: FWIW microsoft docs say the preferred way is `services.AddHttpContextAccessor();`.

Hippo.Web/ClientApp/src/components/ClusterAdmin/Clusters.tsx
sprucely: I tried to get it to work by making the cluster segment of the route optional `pattern: "/api/{cluster?}/{controller=Account}/{action=Index}/{id?}"`, but it would always end up returning the Home view instead of hitting the api action.

Hippo.Web/Controllers/ClusterAdminController.cs
sprucely: `Where()` would filter to a subset, but not `Select()`

Hippo.Web/ClientApp/src/components/ClusterAdmin/Clusters.tsx
srkirkland: oh funny.  alright good enough

srkirkland (approved):

PR #151, opened 2023-01-11
Make db initialization optional

sprucely:


jSylvestre (approved):
Didn't test, but looks good to me

srkirkland (approved):
interesting that quoting the booleans works in the appsettings.  i suppose it makes sense they would convert but still.

PR #164, opened 2023-01-19
Use User.IsAdmin instead of hardcoded values

sprucely:


jSylvestre (approved):
Didn't test, but looks good

PR #173, opened 2023-01-20
Convert to base64 before save

jSylvestre:


jSylvestre:
shouldn't need this. Just don't convert from base64

PR #184, opened 2023-06-05
Approve requests for existing accounts

tefarley:
Automatically approve requests for existing accounts, do not notify the sponsor, and update SSH keys.


srkirkland (approved):
I think this will dump a new file with the user's kerb.txt even though that user will already be created, so puppet just needs to know how to handle that case.  Code itself looks good to me 👍

srkirkland (approved):
I think this will dump a new file with the user's kerb.txt even though that user will already be created, so puppet just needs to know how to handle that case.  Code itself looks good to me 👍

tefarley:
Yup, puppet knows how to merge those conflicts👍

Issue #189, opened 2023-06-08
Use a YAML Serializer for Producing Account File


camillescottatwork:
Currently, the generated YAML is produced manually, here: https://github.com/ucdavis/hippo/blob/main/Hippo.Web/Services/YamlService.cs#L25

This results in some edge cases. For example, a user submitted an SSH key with a comment like so:

    ...ZlRHGWCOKfw8raq0NV9S9LY0T71UUVhJc= ad3\cpw@CAES-FC-AHZL3

Because `account::key` is currently just string-interpolated as `key: "ssh-rsa ..."`, the `\c` is interpreted by YAML/JSON parsers as an escape sequence, and the YAML document is invalid. A YAML serializer would have produced a proper YAML document either by properly escaping the string or leaving out the double quotes that make `key` a string literal and subject to escape codes.

The fix is to either use a YAML library or leave out the double quotes from the `key` field. The former is the better solution, as it will avoid other as-of-yet undiscovered edge cases.

PR #192, opened 2023-06-14
rewritten YamlService.cs with YamlDotNet yaml serialization and questionable anonymous type objects.

tefarley:


Hippo.Web/Services/YamlService.cs
```diff
@@ -26,24 +27,32 @@ public async Task<string> Get(User currentUser, AccountCreateModel accountCreate
...
+                        cluster = sponsorAccount.Cluster
+                    },
+                    account = new
+                    {
+                        name = currentUser.Name,
+                        email = currentUser.Email,
+                        kerb = currentUser.Kerberos,
+                        iam = currentUser.Iam,
+                        mothra = currentUser.MothraId,
+                        key = accountCreateModel.SshKey
```
jSylvestre, line 52: Does it need the extra quotes?

tefarley: pardon? I'm not sure I understand

jSylvestre: The code that was deleted looked like this
sb.AppendLine($"    key: \"{accountCreateModel.SshKey}\"");

With escaped double quotes. I don't remember why we did that originally.

tefarley: Ah I see. You could be right, something else may expect the quotes. Thank you. Lemme check with Omen to be sure it's not a requirement on the puppet side. I'll see if I can do a more in-depth test on my local hippo. If anything I can just add the quotes back in the new version

jSylvestre (approved):

Hippo.Web/Services/YamlService.cs
camillescottatwork: @jSylvestre see #189: you don't want to be in the business of formatting the YAML yourself; the serializer will handle escaping and quoting as-needed automagically.

Hippo.Web/Hippo.Web.csproj
```diff
@@ -21,11 +21,11 @@
     </PackageReference>
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="6.0.2" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="6.0.2" />
-    <PackageReference Include="MvcReact" Version="1.0.1" />
```
jSylvestre, line 24: You probably need to add this back.

sprucely (changes requested):

Hippo.Web/Hippo.Web.csproj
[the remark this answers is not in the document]
tefarley: Ah that was an accident thank you.

sprucely (approved):

PR #198, opened 2023-07-13
Initial implementation of Hippo.Jobs.PuppetSync

sprucely:
Doesn't actually do anything right now. Though it will pull data if you tweak `var accounts = await puppetService.GetPuppetDetails("tbd...");` to actually point to a yaml file in the accounts repo. I didn't want to hard-code that. I'm thinking the yaml file paths should probably be stored in the `dbo.Clusters` table, but wanted to get an alternative App Service running, along with its own db so that deployments from merges to `main` and `dev` don't have migrations clobbering each other. So this should automatically deploy to https://hippo-dev-test.azurewebsites.net

### DIFF
--- a/Hippo.Web/Hippo.Web.csproj
+++ b/Hippo.Web/Hippo.Web.csproj
@@ -26,7 +26,6 @@
     <PackageReference Include="Serilog.Enrichers.ClientInfo" Version="1.1.4" />
     <PackageReference Include="Serilog.Exceptions" Version="8.0.0" />
     <PackageReference Include="Serilog.Sinks.Elasticsearch" Version="8.4.1" />
-    <PackageReference Include="YamlDotNet" Version="13.1.0" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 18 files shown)


Hippo.Web/Hippo.Web.csproj
```diff
@@ -26,7 +26,6 @@
     <PackageReference Include="Serilog.Enrichers.ClientInfo" Version="1.1.4" />
     <PackageReference Include="Serilog.Exceptions" Version="8.0.0" />
     <PackageReference Include="Serilog.Sinks.Elasticsearch" Version="8.4.1" />
-    <PackageReference Include="YamlDotNet" Version="13.1.0" />
```
jSylvestre, line 29: Did you mean to remove it from here?

sprucely: Yes. It's now in Hippo.Core. The versions conflicted, so it seemed cleaner to have it specified in one location rather than trying to keep them in sync.

jSylvestre (approved):
Didn't test, but looks ok to me.

PR #204, opened 2023-07-27
Restructure for Groups and role perms

sprucely:
If you want to test this, be sure to update the connection string in your secrets to point to the correct database. It's the same as the hippo-test connection string, except the database name is `hippo-dev-test`.

The concept of sponsors is replaced with groups. Anyone with GroupAdmin (for a given group), ClusterAdmin (for given cluster) or System permission can approve account requests, though I believe email notifications about requests should only go out to GroupAdmins.

I removed the admin override stuff. If needed, when someone who is not a GroupAdmin approves or rejects a request, we could have it recorded in history as an admin override.

I left the recursive relationship `Account.SponsorId` in place because the `PuppetSyncService` makes use of it when populating the initial set of Groups. After deployment to production, another migration can remove it.

There's still more to be done before `dev` is ready for production (add a `manage groups` tab, merge history tables, update mjml template system, maybe use DataTables) but this PR is already huge.


jSylvestre (approved):
Didn't test, and a lot of changes.
But didn't see any issues

PR #205, opened 2023-08-01
Set toast duration to 30s, updated notification text to say: wait 1-2 days for changes to take.

tefarley:


sprucely (approved):
Looks fine. Any idea why `package-lock.json` updated?

PR #215, opened 2023-08-22
Add Groups Tab

sprucely:


Hippo.Core/Services/AccountSyncService.cs
```diff
@@ -118,7 +118,7 @@ private async Task AddNewGroupMemberships(Cluster cluster)
             // check for new group memberships
             var addPerms = await _dbContext.PuppetGroupsPuppetUsers
                 // only sync users that are members of an existing group
-                .Where(pgpu => pgpu.ClusterName == cluster.Name && _dbContext.Groups.Any(group => group.Name == pgpu.GroupName && group.ClusterId == cluster.Id))
+                .Where(pgpu => pgpu.ClusterName == cluster.Name && _dbContext.Groups.Any(group => group.IsActive && group.Name == pgpu.GroupName && group.ClusterId == cluster.Id))
```
jSylvestre, line 121: If you are never going to look at inactive groups, you could also add it in the query filter

jSylvestre (approved):
Didn't test, but looks good.

PR #216, opened 2023-08-22
Added a button for users to update their ssh key

tefarley:


jSylvestre (changes requested):
If it is an internal link, I believe there is a Link tsx.
Anyway, having the hard coded hostname is bad.

sprucely:
`<NavLink />` is what you're looking for.
https://github.com/ucdavis/hippo/blob/5a21ab2d07727f34b503a46284accc71733345eb/Hippo.Web/ClientApp/src/AppNav.tsx#L39-L48

tefarley:
Who wrote this crap? Oh wait.. ;) I faintly remember trying the navlink before, but the <Link> from router dom is interesting I thought that was only added in Next.js . I'll make some changes

jSylvestre:
I see you have a nullable on the match? Perhaps just hide this if that is null?

tefarley:
I think I accidentally left that in there ty. I can't think hippo rn, where can I get cluster name from in AccountInfo?

tefarley:
nvm ty

tefarley:
Yes ty a nullable match is not great. I can't think hippo well atm where can I get cluster.name from?

PR #219, opened 2023-08-30
TEF/SshKeyRegex

tefarley:
**Temporary Workaround** regex for ssh keys

### DIFF
--- a/Hippo.Web/Controllers/AccountController.cs
+++ b/Hippo.Web/Controllers/AccountController.cs
@@ -142,7 +142,7 @@ public class AccountController : SuperController
     public async Task<ActionResult> Create([FromBody] AccountCreateModel model)
     {
         var currentUser = await _userService.GetCurrentUser();
-        model.SshKey = Regex.Replace(model.SshKey, @"(?<!ssh-rsa)\s+", "");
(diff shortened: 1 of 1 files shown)


Hippo.Web/Controllers/AccountController.cs
```diff
@@ -142,7 +142,7 @@ public async Task<ActionResult> Reject(int id, [FromBody] RequestRejectionModel
     public async Task<ActionResult> Create([FromBody] AccountCreateModel model)
     {
         var currentUser = await _userService.GetCurrentUser();
-        model.SshKey = Regex.Replace(model.SshKey, @"(?<!ssh-rsa)\s+", "");
+        model.SshKey = Regex.Replace(model.SshKey, @"(?<!ssh-rsa)\s+(([\w\.\-]+)@([\w\-]+\.?)+)?", "");
```
jSylvestre, line 145: If they have junk in their SSH, wouldn't it be better to validate that instead of replace it?

sprucely: It's optional metadata. Removing the whitespace ended up concatenating it to the actual key. So this just strips the metadata out.

sprucely (approved):

Hippo.Web/Controllers/AccountController.cs
tefarley: you are correct anyway, but this is just temporary while I am working on a 2 for 1 solution for validation and formatting

PR #220, opened 2023-08-30
Make groups, accounts relationship many-to-many

sprucely:
I'm still limiting the account request page to selecting a single group, since there's nothing in place to support handling approvals for multiple groups.

I removed the `GroupMember` role, since that is now implied by the Accounts->Groups many-to-many relationship.

I extended the AccountSyncService to create missing accounts. Before it was only creating missing users. But now an account is needed to link them to a group.


jSylvestre (approved):
Didn't test, but I don't see any problems

PR #221, opened 2023-09-01
Sync all groups and simplify account sync queries

sprucely:
And remove group delete button, since that is now done by `AccountSyncService`.

### DIFF
--- a/Hippo.Core/Services/AccountSyncService.cs
+++ b/Hippo.Core/Services/AccountSyncService.cs
@@ -28,6 +28,8 @@ namespace Hippo.Core.Services
 
         public async Task SyncAccounts()
         {
+            using var transaction = await _dbContext.Database.BeginTransactionAsync();
+            
             // clear temp data in db
             await _dbContext.TruncateAsync<PuppetGroupPuppetUser>();
 
@@ -38,6 +40,8 @@ namespace Hippo.Core.Services
             {
                 await SyncAccounts(cluster);
             }
+
+            await transaction.CommitAsync();
         }
 
         private async Task SyncAccounts(Cluster cluster)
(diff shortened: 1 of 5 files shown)


Hippo.Core/Services/AccountSyncService.cs
```diff
@@ -119,19 +197,15 @@ private async Task AddNewGroupAccounts(Cluster cluster)
...
-                // identify missing GroupAccounts (Left Join is via GroupJoin/SelectMany)
-                .GroupJoin(_dbContext.GroupsAccounts,
-                    x => new { AccountId = x.account.Id, GroupId = x.group.Id },
-                    ga => new { ga.AccountId, ga.GroupId },
-                    (x, groupAccounts) => new { x.pgpu, x.account, g = x.group, groupAccounts })
-                .SelectMany(x => x.groupAccounts.DefaultIfEmpty(), (x, groupAccount) => new { x.pgpu, x.account, x.g, groupAccount })
-                .Where(x => x.groupAccount == null)
-                .Select(x => new GroupAccount { GroupId = x.g.Id, AccountId = x.account.Id })
+                // identify missing GroupAccounts
+                .Where(x => !_dbContext.GroupsAccounts.Any(ga => ga.AccountId == x.account.Id && ga.GroupId == x.group.Id))
```
jSylvestre, line 201: Worthwhile/possible doing an await for this?

jSylvestre (approved):

Hippo.Core/Services/AccountSyncService.cs
sprucely: Nope. It's part of the linq expression on `_dbContext.PuppetGroupsPuppetUsers` and translated to a query as a whole.

PR #222, opened 2023-09-05
Remove group create form

sprucely:
It's no longer necessary, since all groups are now being synced by the `AccountSyncService`


jSylvestre (approved):
Didn't test.

PR #224, opened 2023-09-06
use react-table

sprucely:
I replaced react-data-table-component with react-table, using PEAKS as an example. A bit more complex, but much closer to our other apps with DataTables.

image


jSylvestre (dismissed):
I was hoping for something that looked more like the original datatables....
Oh well, 🚢

jSylvestre (approved):
🤷‍♂️

PR #226, opened 2023-09-13
Only allow updates of active accounts

### LINKED ISSUE #225: Users can effectively approve their own accounts
It appears that resubmitting an account create request will result in a yaml file being delivered to puppet. I think the intent was to allow users to update their ssh key if they already have an active account. But not enough checks are in place. 

https://github.com/ucdavis/hippo/blob/d31bdf512a9bb42429b4321078de596e5bbae753/Hippo.Web/Controllers/AccountController.cs#L171-L192

sprucely:
Closes #225 

There are also some problems with the information being logged in `AccountController.Create()`, but that's not as urgent, requires more extensive changes, and is already being addressed in the `dev` branch.


jSylvestre (approved):
Looks good.

PR #227, opened 2023-09-14
Merge history tables and refactor HistoryService

sprucely:
Also tweaked some controller actions to be a little more consistent WRT admin overrides.


Hippo.Web/Controllers/AccountController.cs
```diff
@@ -221,12 +233,12 @@ public async Task<ActionResult> Create([FromBody] AccountCreateModel model)
 
         if (existingAccount != null && existingAccount.Status == Account.Statuses.Active)
         {
-            existingAccount.AccountYaml = await _yamlService.Get(currentUser, model, cluster);
-
-            await _historyService.AccountApproved(existingAccount);
-            await _historyService.AddHistory("Existing account override approve", $"Kerb: {existingAccount.Owner.Kerberos} IAM: {existingAccount.Owner.Iam} Email: {existingAccount.Owner.Email} Name: {existingAccount.Owner.Name}", existingAccount.ClusterId, existingAccount.Id);
+            if (existingAccount.Status != Account.Statuses.Active)
```
jSylvestre, line 236: This can't be hit now?
Do we need/want to allow them to update their SSH before it is approved?

jSylvestre (approved):
Didn't test, but except for my one comment, looks good to me.

Hippo.Web/Controllers/AccountController.cs
sprucely: Oops. Forgot to change that above condition.

I believe the intent is that, if an account has already been approved, a user should be able to update their ssh without requiring further approval.

laholstege (approved):
didnt see anything beyond what jason pointed out 👍

PR #234, opened 2023-09-26
Allow any  user to pull available groups.

sprucely:


jSylvestre (approved):
If you don't want to add the auth to UntrackedGroups, then looks ok to me.

PR #248, opened 2023-10-25
Replaced textarea in RequestForm with input and removed regexs in AccountController until I can implement ssh key validation front and back.

tefarley:
Temporary changes to keep non rsa keys being eaten in meantime, until I add validation. Or maybe it won't be temporary depending if everyone likes


sprucely:
Closing, since ssh validation has since been improved in main.

PR #250, opened 2023-11-02
Refactor account sync and group perms

### LINKED ISSUE #239: Account sync needs to detect name/email changes and update accordingly

### LINKED ISSUE #240: Determining what group members are admins for a given group
Once the puppet schema includes a list of group admins/sponsors, the account sync needs to make use of it for setting group-level permissions.

### LINKED ISSUE #241: We don't need puppet sync to create a user record for every account; just for those accounts that are group admins/sponsors. Add Account Email field and make OwnerId optional.

### LINKED ISSUE #242: Let EF Core abstract/hide the GroupsAccounts table

sprucely:
The EF Core Bulk Extensions wasn't as flexible as I had hoped. But the sync code is still much simpler now.

Closes #241
Closes #239
Closes #240
Closes #242


Hippo.Core/Domain/Request.cs
```diff
@@ -17,10 +17,8 @@ public class Request
         public int RequesterId { get; set; }
         public User Actor { get; set; }
         public int? ActorId { get; set; }
-        public Group Group { get; set; }
-        public int? GroupId { get; set; }
-        public Account Account { get; set; }
-        public int? AccountId { get; set; }
+        [MaxLength(32)]
+        public string Group { get; set; }
```
srkirkland, line 21: if the sync happens and a group gets deleted, won't this prevent that?  i'm guessing it would throw an FK error when trying to delete and stop the whole bulk delete?

Hippo.Core/Domain/Group.cs
```diff
@@ -16,34 +16,36 @@ public class Group
...
         public int ClusterId { get; set; }
         [Required]
         public Cluster Cluster { get; set; }
 
-        public List<Permission> Permissions { get; set; } = new();
+        [JsonIgnore]
+        public List<Account> MemberAccounts { get; set; } = new();
+
+        [JsonIgnore]
+        public List<GroupMemberAccount> MemberAccountsJunction { get; set; } = new();
```
srkirkland, line 28: What is this `MemberAccountsJunction` and `AdminAccountsJunction` needed?  i don't see it used in mapping or in code anywhere, and it seems like the existing `MemberAccounts` and `AdminAccounts` lists should handle what is needed, right?

srkirkland (dismissed):

Hippo.Core/Domain/Group.cs
sprucely: Oh, yeah I can delete those. Ended up referencing the junction DbSets directly from AppDbContext.

Hippo.Core/Domain/Request.cs
sprucely: Nope. It's not a foreign key. I probably should have named it `GroupName`. Queries referencing it have been updated to perform a left join to get at the associated group.

PR #251, opened 2023-11-06
Fix cluster names in notifications and use Mjml.Net

### LINKED ISSUE #236: Can we have a 'friendly' name for the clusters? So we don't send "Peloton is a Linux Based Supercomputer primarily used by the UC Davis College of Letters and Science account" and we just send Peloton. Or Farm. Or Franklin.

sprucely:
Closes #236


srkirkland (approved):

jSylvestre (approved):
Looks good to me.

PR #252, opened 2023-11-07
Improve ssh key handling and validation

### LINKED ISSUE #183: Validate user uploaded SSH public keys
It would be nice to do as much validation of the SSH public key that users upload as possible. There is a good description of the format the the base64 encoded portion here https://www.thedigitalcatonline.com/blog/2018/04/25/rsa-keys/ under the `The OpenSSH public key format` section. The very short version is the first 4 bytes represent the length of the next chunk. The 2nd chunk is `ssh-` and the algorithm. The first part of each of my 3 public keys look like this in hex:
```
0000000 00 00 00 07 73 73 68 2d 72 73 61 00 00 00 03 01

0000000 00 00 00 0b 73 73 68 2d 65 64 32 35 35 31 39 00

0000000 00 00 00 07 73 73 68 2d 72 73 61 00 00 00 03 01
```

And like this in ascii: 
```
0000000 nul nul nul bel   s   s   h   -   r   s   a nul nul nul etx soh

0000000 nul nul nul  vt   s   s   h   -   e   d   2   5   5   1   9 nul

0000000 nul nul nul bel   s   s   h   -   r   s   a nul nul nul etx soh
```

So very basic validation would be that bytes 0-3 are a small integer, and that 4-7  are `73 73 68 2d`.

### LINKED ISSUE #203: SSH public keys need to be a single line
Users have been known to submit multi-line SSH public keys, which do not work. An easy fix before #183 is implemented would be to make the SSH input box only accept a single line of input. Additionally, all trailing white space needs to be stripped.

sprucely:
Not going as far as loading any encryption libraries to parse the ssh key. I think it's good enough just to ensure the base-64 encoded header matches the declared key type.

Closes #183
Closes #203

### DIFF
--- a/Hippo.Core/Extensions/StringExtensions.cs
+++ b/Hippo.Core/Extensions/StringExtensions.cs
@@ -1,3 +1,4 @@
+using System.Text;
 using System.Text.RegularExpressions;
 
 public static class StringExtensions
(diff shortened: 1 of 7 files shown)


jSylvestre (approved):
If the regex is correct...

PR #256, opened 2023-11-08
Fix a HippoV2 migration bug and account sync bug

### LINKED ISSUE #254: Determine why cluster admin perms weren't created in migration

sprucely:
Closes #254


srkirkland (approved):
cool -- don't think i've done custom migration sql before but looks good to me.

PR #258, opened 2023-11-13
Make all group columns sortable and filterable

### LINKED ISSUE #247: Fix group column filtering

sprucely:
I applied the accessor/Cell fix to the three remaining table columns that contain the `<GroupNameWithTooltip />` component. Also updated the accessors to include more details from `GroupModel` for better filtering.

Closes #247

### DIFF
--- a/Hippo.Web/ClientApp/src/util/StringHelpers.ts
+++ b/Hippo.Web/ClientApp/src/util/StringHelpers.ts
@@ -1,3 +1,5 @@
+import { GroupModel } from "../types";
+
 export const SplitCamelCase = (str: string) => {
   return (
     str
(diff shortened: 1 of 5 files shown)


Hippo.Web/ClientApp/src/util/StringHelpers.ts
```diff
@@ -7,3 +9,17 @@ export const SplitCamelCase = (str: string) => {
...
+
+export const getGroupModelString = (arg: GroupModel | GroupModel[]) => {
+  const groups = Array.isArray(arg) ? arg : [arg];
+  var values = [
+    // start with names of groups for sortability
+    ...groups.map((g) => g.name),
+    // then include remaining properties for filtering
+    ...groups.map((g) => [
+      g.displayName,
+      g.admins.map((a) => [a.name, a.email, a.kerberos]),
```
srkirkland, line 21: seems like it'd be a little confusing to me if you type in the admin's name or email and the row filtered, unless it displayed the admin in the actual cell, which i don't think it does.

sprucely: It displays admin name and email in the tooltip.

PR #264, opened 2023-11-20
A few more fixes for HippoV2

### LINKED ISSUE #261: sort by group name by default
might be worth making sure each table has a simple default sort

sprucely:
Closes #261

### DIFF
--- a/Hippo.Core/Domain/Account.cs
+++ b/Hippo.Core/Domain/Account.cs
@@ -42,8 +42,8 @@ namespace Hippo.Core.Domain
         public int? OwnerId { get; set; }
         public User Owner { get; set; }
 
-        public int ClusterId { get; set; }
         [Required]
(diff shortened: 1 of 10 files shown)


srkirkland (approved):

Hippo.Core/Domain/Account.cs
```diff
@@ -42,8 +42,8 @@ public Account()
         public int? OwnerId { get; set; }
         public User Owner { get; set; }
 
-        public int ClusterId { get; set; }
         [Required]
```
jSylvestre, line 45: Do you want this required on the int, or the Cluster below?

sprucely: It has same impact as far as EF is concerned. However, having it on the navigation property is problematic when doing explicit validations in AccountSyncService, because it's setting just the int and not the navigation property.

PR #265, opened 2023-11-21
Some ssh-key entry tweaks

sprucely:
I moved the example from help text to the placeholder, so that it can't be easily copied. Also added link to relevant ssh documentation.

image

### DIFF
--- a/Hippo.Web/ClientApp/src/components/Account/AccountInfo.tsx
+++ b/Hippo.Web/ClientApp/src/components/Account/AccountInfo.tsx
@@ -107,11 +107,10 @@ export const AccountInfo = () => {
                   required
                   onChange={(e) => {
                     const value = e.target.value
-                      .trim()
                       .replaceAll("\r", "")
                       .replaceAll("\n", "");
                     e.target.value = value;
-                    setReturn(e.target.value);
+                    setReturn(e.target.value.trim());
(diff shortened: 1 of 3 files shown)


Hippo.Web/ClientApp/src/components/Account/AccountInfo.tsx
```diff
@@ -103,23 +103,24 @@ export const AccountInfo = () => {
...
-                  <br></br>
-                  <code>
-                    ssh-rsa AAAAB3NzaC1yc....NrRFi9wrf+M7Q== fake@addr.local
-                  </code>
+                  For more information on SSH keys, please see{" "}
+                  <a
+                    href="https://wiki.cse.ucdavis.edu/support:general:security:ssh"
+                    target={"blank"}
+                  >
+                    https://wiki.cse.ucdavis.edu/support:general:security:ssh
```
jSylvestre, line 122: Want to use a config value for this in case it changes?

srkirkland (dismissed):
@sprucely going to approve, but that whole request form is basically duplicated twice, it should probably just go ahead and be its own component, especially since they seem to be picky about changing the ssh stuff.

srkirkland:
Also make sure to link/close the related issue @sprucely

PR #275, opened 2023-12-14
Replace personal access token with github app authentication 

sprucely:
I couldn't use Octokit's suggestion to use the GitHubJwt nuget package for generating the jwt, because BouncyCastle would fail to read the pem key no matter how I formatted it. And just about all the examples out there are based on BouncyCastle. But it turns out to be not too complex just using `System.Security.Cryptography` directly.

I've updated secrets in 1pass.

### DIFF
--- a/Hippo.Core/Hippo.Core.csproj
+++ b/Hippo.Core/Hippo.Core.csproj
@@ -22,7 +22,7 @@
     <PackageReference Include="Mjml.Net" Version="3.3.0" />
     <PackageReference Include="Octokit" Version="7.0.0" />
 		<PackageReference Include="Serilog.AspNetCore" Version="4.1.0" />
-		<PackageReference Include="SSH.NET" Version="2020.0.1" />
+		<PackageReference Include="SSH.NET" Version="2023.0.0" />
 		<PackageReference Include="YamlDotNet" Version="13.1.1" />
 	</ItemGroup>
 
--- a/Hippo.Core/Services/PuppetService.cs
+++ b/Hippo.Core/Services/PuppetService.cs
@@ -30,7 +30,7 @@ namespace Hippo.Core.Services
         private async Task<GitHubClient> GetGithubClient()
         {
             // caching token for 10 minutes, which is the limit for GitHub API's
-            if (!_memoryCache.TryGetValue("github-app-installation-token", out string appInstallationToken))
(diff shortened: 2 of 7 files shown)


Hippo.Core/Hippo.Core.csproj
```diff
@@ -22,7 +22,7 @@
     <PackageReference Include="Mjml.Net" Version="3.3.0" />
     <PackageReference Include="Octokit" Version="7.0.0" />
 		<PackageReference Include="Serilog.AspNetCore" Version="4.1.0" />
-		<PackageReference Include="SSH.NET" Version="2020.0.1" />
+		<PackageReference Include="SSH.NET" Version="2023.0.0" />
```
srkirkland, line 25: ssh.net downgrade?

jSylvestre (approved):
Didn't test, but looks ok.

Hippo.Core/Services/PuppetService.cs
```diff
@@ -17,23 +19,55 @@ public interface IPuppetService
...
         {
             _settings = settings.Value;
-            _gitHubClient = new GitHubClient(new ProductHeaderValue("Hippo"))
+            _memoryCache = memoryCache;
+        }
+
+        private async Task<GitHubClient> GetGithubClient()
+        {
+            // caching token for 10 minutes, which is the limit for GitHub API's
+            if (!_memoryCache.TryGetValue("github-app-installation-token", out string appInstallationToken))
```
srkirkland, line 33: usually it's safer or at least more common to use `GetOrCreateAsync<T>` when working with memory cache since it'll handle the missing data case and wraps it nicely so you don't have to worry about race conditions, plus just more obvious what it is trying to do

srkirkland (changes requested):

Hippo.Core/Hippo.Core.csproj
sprucely: upgrade 2020 to 2023

Hippo.Core/Services/PuppetService.cs
sprucely: GetOrCreate doesn't appear to have an override that takes `MemoryCacheEntryOptions` arg.

sprucely: Oh, I see. The delegate provides a `ICacheEntry` arg. Okay.

PR #278, opened 2024-01-10
Fix request completion and don't use explicit transaction

sprucely:


Hippo.Core/Services/AccountSyncService.cs
```diff
@@ -27,168 +28,172 @@ public AccountSyncService(IPuppetService puppetService, AppDbContext dbContext,
...
-            using var transaction = await _dbContext.Database.BeginTransactionAsync();
+            bool success = true;
 
             foreach (var cluster in await _dbContext.Clusters
                 .Where(c => !string.IsNullOrEmpty(c.Domain))
                 .AsNoTracking()
                 .ToArrayAsync())
             {
-                await Run(cluster);
+                success &= await Run(cluster);
```
srkirkland, line 40: wtf

srkirkland (approved):

PR #282, opened 2024-01-17
Add cluster-specific emails for help content and links

sprucely:


Hippo.Core/Domain/Cluster.cs
```diff
@@ -29,6 +29,8 @@ public class Cluster
         public bool IsActive { get; set; } = true;
         [MaxLength(250)]
         public string Domain { get; set; } = String.Empty;
+        [MaxLength(250)]
```
jSylvestre, line 32: I think you can add an email attribute here. Don't know if it is worth it.

jSylvestre (dismissed):

PR #287, opened 2024-01-19
Fix EF Core warnings about global query filters

### LINKED ISSUE #283: Fix global query filter warnings

sprucely:
Closes #283


srkirkland (commented):
Just be careful and double check everything. With these multi-level query filters it's easy to get mixed up and filter more than you expect.

PR #288, opened 2024-01-22
Add user ssh key option, simplify cluster admin DataTable, and add cluster details modal

### LINKED ISSUE #238: Make SSH keys optional on a per-cluster basis
The new CF cluster is likely to use campus passphrases instead of SSH keys, so we need a way, on a per-cluster basis, to not require, and perhaps not even allow, SSH key submission.

sprucely:
Copilot's suggestion for a readonly version of the cluster edit form basically amounted copying it and applying readOnly attributes to all the inputs. But it seemed good enough to me.

Closes #238


Hippo.Web/ClientApp/src/components/Account/RequestForm.tsx
```diff
@@ -40,10 +43,10 @@ export const RequestForm = () => {
     };
 
     fetchGroups();
-  }, [cluster]);
+  }, [clusterName]);
 
   const handleSubmit = async () => {
-    const req = authenticatedFetch(`/api/${cluster}/account/create`, {
+    const req = authenticatedFetch(`/api/${clusterName}/account/create`, {
```
jSylvestre, line 49: Can the cluster name contain invalid characters?
Maybe use a cluster slug like we do in some of our other team apps?

srkirkland:
> Copilot's suggestion for a readonly version of the cluster edit form basically amounted copying it and applying readOnly attributes to all the inputs. But it seemed good enough to me.
> 
> Closes #238

ha lazy copilot

Hippo.Web/ClientApp/src/components/Account/RequestForm.tsx
sprucely: It's only a variable name change, since `cluster` is now an object of type `Cluster`.

jSylvestre (approved):

PR #295, opened 2024-02-02
Replace yaml uploads with EventQueue api

### LINKED ISSUE #229: REST API for querying users whose accounts need to be added
Instead of having Hippo push YAML files, we can turn it around so Hippo has a REST API that can be queried for pending account creations. Once the accounts are created, we can POST to the API so Hippo knows the process finished successfully on the cluster.

sprucely:
The requested changes are automatically applied when a `QueuedEvent.Status` is set to `Complete`. With the account sync still running hourly, it's possible for those changes to get temporarily rolled back if the change in question has not yet made it to the puppet repo's main branch.

I copied the `Result` class from Harvest. It combines return values, messages, is-success flag and logging in a handy little api.

I started to add an AccountStatusChanged email to provide notifications about when a change is completed, but decided I should probably wait for the feature request. But I left in the fix for those `Decission` misspellings that have been bugging me.

Closes #229

Edit: The old behavior of sending updates to puppet via yaml is still enabled by default. To switch over to usage of EventQueue, set `"EventQueueEnabled": true` in user secrets.


Hippo.Core/Domain/QueuedEvent.cs
```diff
@@ -0,0 +1,64 @@
...
+    public int Id { get; set; }
+    [MaxLength(50)]
+    public string Action { get; set; } = "";
+    [MaxLength(50)]
+    public string Status { get; set; } = "";
+    public string Data { get; set; } = "";
+    public string ErrorMessage { get; set; } = "";
+    public DateTime CreatedAt { get; set; } = DateTime.UtcNow;
+    public DateTime UpdatedAt { get; set; } = DateTime.UtcNow;
+    public int? RequestId { get; set; }
```
srkirkland, line 17: how can you have a queued event that doesn't have an attached request?

Hippo.Core/Domain/QueuedEvent.cs
```diff
@@ -0,0 +1,64 @@
...
+    public Request Request { get; set; }
+
+    internal static void OnModelCreating(ModelBuilder modelBuilder)
+    {
+        modelBuilder.Entity<QueuedEvent>().HasIndex(qe => qe.Action);
+        modelBuilder.Entity<QueuedEvent>().HasIndex(qe => qe.Status);
+        modelBuilder.Entity<QueuedEvent>().HasIndex(qe => qe.CreatedAt);
+        modelBuilder.Entity<QueuedEvent>().HasIndex(qe => qe.UpdatedAt);
+        modelBuilder.Entity<QueuedEvent>().HasIndex(qe => qe.RequestId);
```
srkirkland, line 26: i don't think indexes are needed on FKs, it's automatic in SQL server

Hippo.Web/Handlers/ApiKeyAuthFilter.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    private readonly IConfiguration _configuration;
+    public ApiKeyAuthFilter(IConfiguration configuration)
+    {
+        _configuration = configuration;
+    }
+
+    public void OnAuthorization(AuthorizationFilterContext context)
+    {
+        string userApiKey = context.HttpContext.Request.Headers[ApiKeyConstants.ApiKeyHeaderName].ToString();
```
srkirkland, line 23: careful of NRE if header doesn't exist

Hippo.Core/Domain/QueuedEvent.cs
sprucely: Requests are for actions that need to be approved. There's no request associated with a user updating their ssh key.

PR #316, opened 2024-04-03
AccesTypes and nested JSON data refactor

### LINKED ISSUE #299: Allow Open OnDemand users to not submit SSH keys
Open OnDemand users auth via CAS through a web site so technically they do not require SSH keys. We may need to add some flexibility to Hippo. My first thought would be to add a drop down with a list like this

What kind of access are you requesting:
* Open OnDemand (web only)
* sbatch/srun/interactive-terminal (SSH to head node)
* Both <- DEFAULT

The 1st option would not require a SSH key, the other 2 would. 

This may create an increased support burden on us when users are confused between the options and don't submit a key when they actually need SSH access. On the other hand, it would prevent users who only need Open OnDemand access from even knowing about SSH keys.

sprucely:
Sorry for another big one. Adding AccessTypes to Cluster, Account and related models, and moving some properties into a nested json field ended up touching just about everything.

The migrations execute some raw sql to move data around. I tested both directions to make sure no data is lost. Originally wasn't planning to do a restore of data from json when reverting the OpenOnDemand migration. That's why I marked a couple columns `Obsolete` instead of deleting them. Copying the data out of json turned out pretty simple, but I didn't want to go through rescaffolding the migration again, so those properties can be removed in a future PR.

I tried to find a way to make the `Data` property on C# classes strongly typed instead of a json string. But csharp's lack of discriminated unions made it too messy. Worked well on the TS side though, parsing the data property upon pulling it from the api.

image

image

image

closes #299


Hippo.Web/ClientApp/src/components/Account/Requests.tsx
```diff
@@ -26,7 +27,12 @@ export const Requests = () => {
       );
 
       if (response.ok) {
-        setRequests(await response.json());
+        setRequests(
+          (await response.json()).map((r) => ({
+            ...r,
+            data: r.data && JSON.parse(r.data),
+          })),
+        );
```
srkirkland, line 35: replace w/ helper to translate response to one of these request objects

PR #318, opened 2024-04-15
Sync and display puppet resource data

sprucely:
I went with a generic collapsible `ObjectTree` component for displaying resource data. Not sure if @cydoval should have a go yet at prettying it up. Maybe @acgetchell will have some feedback?

Group data:
image

Account data:
image


jSylvestre (approved):
Didn't test, but looks ok.

PR #321, opened 2024-04-30
Jcs/ordering system

jSylvestre:


Hippo.Core/Domain/Order.cs
```diff
@@ -0,0 +1,104 @@
...
+        [Required]
+        [MaxLength(50)]
+        public string Name { get; set; }
+        [MaxLength(250)]
+        public string Description { get; set; }
+        [MaxLength(150)]
+        public string ExternalReference { get; set; }
+        [Required]
+        [Range(0.01, double.MaxValue)]
+        public decimal Price { get; set; }
```
srkirkland, line 28: we'll need "units" and then probably change this to "UnitPrice".  Units would be like "TB" for a terabyte of storage, "Point" for a fair share point, etc.  It's just for display purposes and can default to something generic, but I think we'll want it

Hippo.Core/Domain/Order.cs
```diff
@@ -0,0 +1,104 @@
...
+        [Required]
+        [Range(1, int.MaxValue)]
+        public int Installments { get; set; }
+
+
+        public decimal Adjustment { get; set; }
+        public string AdjustmentReason { get; set; }
+        public decimal SubTotal { get; set; }
+        public decimal Total { get; set; }
+        public decimal BalanceRemaining { get; set; } // do here, or just get the total of the billings?
```
srkirkland, line 41: Agree we should do here.  Not against also doing a total for final check, but this is good to keep for display and sanity checks

Hippo.Core/Domain/Order.cs
```diff
@@ -0,0 +1,104 @@
...
+        public string Notes { get; set; }
+        public string AdminNotes { get; set; }
+        public string Status { get; set; }
+
+        public decimal InstalmentAmount => Math.Round(Total / Installments, 2);
```
srkirkland, line 46: In America we use 2 "L"'s in installment 🇨🇦 .

Hippo.Core/Domain/Order.cs
```diff
@@ -0,0 +1,104 @@
...
+        public int ClusterId { get; set; }
+        public Cluster Cluster { get; set; }
+
+        [Required]
+        public int CreatedById { get; set; }
+        public User CreatedBy { get; set; }
+
+        [Required]
+        public int PrincipalInvestigatorId { get; set; }
+        public User PrincipalInvestigator { get; set; }
```
srkirkland, line 58: @sprucely help me here but do we want to tie this to "Account" table or "User" table?  Also I think i'd say change the name to something not so PI-focused.  Like `OrderedFor` or `Submitter` or something.  It's just PIs for now and in farm, but let's not 100% rely on that forever

Hippo.Core/Domain/FinancialDetail.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.ComponentModel.DataAnnotations;
+
+namespace Hippo.Core.Domain
+{
+    public class FinancialDetail
+    {
+        [Key]
+        int Id { get; set; }
+        [StringLength(128)] //Probably doesn't need to be this big...
+        string SlothApiKey { get; set; }
```
srkirkland, line 11: `FinancialSystemApiKey` or similar.  Also we might not hold the actual key here but a pointer to the key secret... but I'm ok leaving the name *ApiKey.  Same thing for sloth source below.

Hippo.Core/Domain/OrderMetaData.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Text.Json.Serialization;
+using System.Threading.Tasks;
+
+namespace Hippo.Core.Domain
+{
+    public class OrderMetaData
+    {
+        [Key]
+        [JsonIgnore]
+        public string Id { get; set; } = Guid.NewGuid().ToString();
```
srkirkland, line 16: why not just use the regular identity id?

Hippo.Core/Domain/Payment.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Hippo.Core.Domain
+{
+    public class Payment
+    {
+        [Key]
+        public int Id { get; set; }
+        public string SlothId { get; set; }
```
srkirkland, line 14: `FinancialSystemId`

Hippo.Core/Domain/Payment.cs
```diff
@@ -0,0 +1,32 @@
...
+        public string TrackingNumber { get; set; } // KFS tracking number
+        public DateTime CreatedOn { get; set; } = DateTime.UtcNow;
+        public decimal Amount { get; set; }
+        public string Status { get; set; }
+
+        public string ClusterChartString { get; set; }
+        public string OrderChartStrings { get; set; } //.Serialize() of the billings chart strings and percentages used
```
srkirkland, line 21: I think you should remove both of these.  Just replace with a single string `Description` field and dump payment details in there.  Like `$5 paid.  $4 from CS1, $1 from CS2, sent to CS3`.  No need to json or anything

Hippo.Core/Domain/Product.cs
```diff
@@ -0,0 +1,33 @@
...
+        [Required]
+        public string Category { get; set; }
+        [Required]
+        [MaxLength(50)]
+        public string Name { get; set; }
+        [MaxLength(250)]
+        public string Description { get; set; }
+        [Required]
+        [Range(0.01, double.MaxValue)]
+        public decimal Price { get; set; }
```
srkirkland, line 20: Like above, I think you'll need "Units" and then this is unitPrice.

Hippo.Core/Domain/Product.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Microsoft.EntityFrameworkCore;
+using System.ComponentModel.DataAnnotations;
+
+namespace Hippo.Core.Domain
+{
+    public class Product
```
srkirkland, line 6: are you going to allow products to be updated? Or just delete/create/clone? If it's just delete/create I think order linking to product is maybe worth it, even if we don't use it in the UI.  Also a create date or last updated data here would be handy

Hippo.Core/Domain/Order.cs
sprucely: I think User is appropriate. Account records can be deleted (not just marked inactive) by puppet sync.

jSylvestre: I was going to have createdBy and PrincipalInvestigator be the same user if the PI initiates the order.
Would only be different if the admin initiates it.
I could also just have one and make a note of that in the history.

Hippo.Core/Domain/OrderMetaData.cs
jSylvestre: Was just copying what sloth did :) 
I'm fine with just an int

Hippo.Core/Domain/Product.cs
jSylvestre: I was going to allow update

Hippo.Core/Domain/Order.cs
srkirkland: Sounds good, I'd still like to make the field name not tied to PI so explicitly.

Hippo.Web/ClientApp/src/components/Order/BillingsFields.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+  const removeBilling = (index: number) => {
+    remove(index);
+  };
+
+  const [notification, setNotification] = useState<string>("");
+  const showNotification = (message: string) => {
+    setNotification(message);
+    // Optionally, clear the notification after some time
+    setTimeout(() => setNotification(""), 5000);
```
srkirkland, line 70: this will get squirrely if this is called >1, since the old setTimeout will still be running and can clear the new notification instantly.  better to either just use toast of if you want to use local, then handle the clearTimeout() logic properly to restart the timeout whenever a new notification is set

Hippo.Web/ClientApp/src/components/Order/BillingsFields.tsx
```diff
@@ -0,0 +1,277 @@
...
+          isValid: rtValue.isValid,
+          description: rtValue.description,
+          accountManager: rtValue.accountManager,
+          accountManagerEmail: rtValue.accountManagerEmail,
+          message: rtValue.message,
+          warning: rtValue.warning,
+        },
+      });
+    } else {
+      alert("Failed!");
```
srkirkland, line 98: !!

Hippo.Web/ClientApp/src/components/Order/ChartStringValidation.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+type ChartStringStatus = "loading" | "error" | "success" | undefined;
+
+const ChartStringValidation: React.FC<ChartStringValidationProps> = ({
+  chartString,
+}) => {
+  const [chartStringValidation, setChartStringValidation] =
+    React.useState<ChartStringValidationModel>(null);
+  const [status, setStatus] = React.useState<ChartStringStatus>(undefined);
```
srkirkland, line 18: undefined is implicit here, you don't need to pass it

Hippo.Web/ClientApp/src/components/Order/ChartStringValidation.tsx
```diff
@@ -0,0 +1,87 @@
...
+      } catch (error) {
+        setStatus("error");
+      }
+    };
+
+    validateChartString();
+  }, [chartString]);
+
+  // default to showing loading state on mount
+  if (status === "loading" || status === undefined) {
```
srkirkland, line 51: if you want to default to loading you can just have the initial `useState("loading")` and get rid of undefined entirely

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { useEffect, useState } from "react";
+import { OrderModel } from "../../types";
+import { useParams } from "react-router-dom";
+import { usePermissions } from "../../Shared/usePermissions";
+import { usePromiseNotification } from "../../util/Notifications";
+import OrderForm from "./OrderForm";
+import { authenticatedFetch, parseBadRequest } from "../../util/api";
+
+const defaultOrder: OrderModel = {
```
srkirkland, line 9: would be nice to have in different file, but can be moved with later refactoring

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+        if (response.ok) {
+          const data = await response.json();
+          setOrder(data);
+          // const balanceRemaining = parseFloat(data.balanceRemaining);
+          // setBalanceRemaining(balanceRemaining);
+          // const balancePending = data?.payments
+          //   .filter((payment) => payment.status !== "Completed")
+          //   .reduce((acc, payment) => acc + parseFloat(payment.amount), 0);
+          // setBalancePending(balancePending);
```
srkirkland, line 63: ✂️

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
```diff
@@ -0,0 +1,175 @@
...
+        } else {
+          alert("Error fetching product for order");
+        }
+      };
+
+      fetchProductOrder();
+    } else {
+      if (isClusterAdmin === false) {
+        window.location.href = `/${cluster}/product/index`;
```
srkirkland, line 72: use the routing APIs, not window.  this would cause a full refresh too

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
```diff
@@ -0,0 +1,175 @@
...
+      } else {
+        setOrder(defaultOrder);
+      }
+    }
+  }, [cluster, isClusterAdmin, productId]);
+
+  // async function so the form can manage the loading state
+  const submitOrder = async (updatedOrder: OrderModel) => {
+    const editedOrder: OrderModel = {
```
srkirkland, line 81: maybe this is just for testing, but it seems like you shouldn't need any of it, just `use the updatedOrder` passed in.

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
```diff
@@ -0,0 +1,175 @@
...
+      adminNotes: updatedOrder.adminNotes,
+      metaData: updatedOrder.metaData,
+      lifeCycle: updatedOrder.lifeCycle,
+      expirationDate: updatedOrder.expirationDate,
+      installmentDate: updatedOrder.installmentDate,
+      billings: updatedOrder.billings,
+    };
+
+    console.log(editedOrder);
+    console.log(JSON.stringify(editedOrder));
```
srkirkland, line 119: ✂️

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
```diff
@@ -0,0 +1,175 @@
...
+        return "An error happened, please try again.";
+      }
+    });
+
+    const response = await req;
+
+    if (response.ok) {
+      const data = await response.json();
+
+      window.location.href = `/${cluster}/order/details/${data.id}`;
```
srkirkland, line 140: use routing

Hippo.Core/Services/PaymentsService.cs
```diff
@@ -0,0 +1,140 @@
...
+                    Log.Information("Skipping order {0} because it has a created or processing payment", order.Id);
+                    continue;
+                }
+                SetNextPaymentDate(order);
+
+                _dbContext.Orders.Update(order);
+                await _dbContext.SaveChangesAsync();
+            }
+            
+            var orders = await _dbContext.Orders.Include(a => a.Payments).Include(a => a.Cluster).Include(a => a.PrincipalInvestigator).Where(a => a.Status == Order.Statuses.Active && a.NextPaymentDate != null && a.NextPaymentDate.Value.Date <= DateTime.UtcNow.Date).ToListAsync();
```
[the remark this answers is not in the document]
jSylvestre: I need Payments, and if I add history values, I'll need the cluster. But I'll remove for now.

Hippo.Web/ClientApp/src/App.tsx
```diff
@@ -19,6 +19,13 @@ import { Clusters } from "./components/Account/Clusters";
 import { Clusters as AdminClusters } from "./components/ClusterAdmin/Clusters";
 import { Groups } from "./components/Admin/Groups";
 import { ShowFor } from "./Shared/ShowFor";
+import { FinancialDetail } from "./components/Financial/FinancialDetail";
+import { Products } from "./components/Product/Products";
+import { Orders } from "./components/Order/Orders";
+import { Details } from "./components/Order/Details";
+import CreateOrder from "./components/Order/CreateOrder";
+import EditOrder from "./components/Order/EditOrder";
+import UpdateChartStrings from "./components/Order/UpdateChartStrings";
```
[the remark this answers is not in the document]
jSylvestre: Will do. Think this happens when they get added automatically and I don't look closer

Hippo.Web/ClientApp/src/components/Order/BillingsFields.tsx
jSylvestre: @srkirkland Not sure how to get toasts to work with that. But... it works fine if it is called quickly multiple times.

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
jSylvestre: Not sure about this one.

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
jSylvestre: @laholstege ☝️  ? I think this was code you created, then I just added to it. (Or maybe I copied it from a different page...)

Hippo.Web/ClientApp/src/components/Order/UpdateChartStrings.tsx
```diff
@@ -0,0 +1,135 @@
...
+      metaData: updatedOrder.metaData,
+      lifeCycle: updatedOrder.lifeCycle,
+      expirationDate: updatedOrder.expirationDate,
+      installmentDate: updatedOrder.installmentDate,
+      billings: updatedOrder.billings,
+    };
+
+    console.log(editedOrder);
+    console.log(JSON.stringify(editedOrder));
+    debugger;
```
laholstege, line 79: 😱

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+  const [updateStatusModel, setUpdateStatusModel] =
+    useState<UpdateOrderStatusModel | null>(null);
+  const adminEditableStatuses = ["Processing", "Active"];
+  const sponsorEditableStatuses = ["Created"];
+
+  useEffect(() => {
+    setIsClusterAdmin(isClusterAdminForCluster());
+  }, [isClusterAdmin, isClusterAdminForCluster]);
+
+  const calculateBalanceRemaining = (data: any) => {
```
srkirkland, line 42: do we need any here?  would be nice to have a real type

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+    const balanceRemaining = parseFloat(data.balanceRemaining);
+    setBalanceRemaining(balanceRemaining);
+    const balancePending = data?.payments
```
srkirkland, line 45: 2 lines earlier data was not nullable, now here it is?

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+
+  useEffect(() => {
+    const fetchOrder = async () => {
+      const response = await authenticatedFetch(
+        `/api/${cluster}/order/get/${orderId}`,
+      );
+
+      if (response.ok) {
+        const data = await response.json();
+        console.log(data);
```
srkirkland, line 62: ✂️

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+        setOrder(data);
+        calculateBalanceRemaining(data);
+      } else {
+        alert("Error fetching order");
```
srkirkland, line 66: use toast/notifications?  Also in general, why sometimes you use the promise notification callback thing and sometimes not?

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+      cell: (value) => (
+        <>
+          {value.row.original.actedBy && (
+            <>
+              {value.row.original.actedBy.firstName}{" "}
+              {value.row.original.actedBy.lastName} (
+              {value.row.original.actedBy.email})
+            </>
+          )}
+          {!value.row.original.actedBy && <>System</>}
```
srkirkland, line 126: clearer if you use a tertiary (`actedBy ? (name) : system`) instead of doing multiple checks against the actedBy value.

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+          )}
+          {!value.row.original.createdBy && <>System</>}
+        </>
+      ),
+    }),
+  ];
+
+  // async function so the form can manage the loading state
+  const submitOrder = async (updatedOrder: OrderModel) => {
+    const editedOrder: OrderModel = {
```
srkirkland, line 178: why do the copy field by field, what's the point of this code?

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+          <div className="form-group">
+            <label htmlFor="fieldAmount">Amount</label>
+            <input
+              className="form-control"
+              id="fieldAmount"
+              required
+              value={editPaymentModel.entryAmount}
+              onChange={(e) => {
+                const value = e.target.value;
+                if (/^\d*\.?\d*$/.test(value) || /^\d*\.$/.test(value)) {
```
srkirkland, line 243: good but you can probably also just use input type number w/ some parameters (step, min/max) to validate in real-time

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+    useConfirmationDialog<UpdateOrderStatusModel>(
+      {
+        title: "Update Order Status",
+        message: (setReturn) => (
+          <>
+            <div>Current Status: {updateStatusModel.currentStatus}</div>
+            <div>Set Status to: {updateStatusModel.newStatus}</div>
+            <hr />
+            {updateStatusModel.newStatus === "Submitted" && (
+              <div style={{ backgroundColor: "#90ee90" }}>
```
srkirkland, line 313: use classes

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+              </div>
+            )}
+          </>
+        ),
+        canConfirm: !notification.pending,
+      },
+      [order, notification.pending, updateStatusModel],
+    );
+
+  const updateStatus = useCallback(async () => {
```
srkirkland, line 336: why useCallback here?  why here and not reject or cancel methods:?

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+          </ShowFor>
+          {/* If you are the sponsor (PI) and it is in the created status, you can move it to submitted if there is billing info */}
+          <ShowFor
+            condition={
+              order.piUser?.id === user.detail.id &&
+              order.status === "Created" &&
+              order.billings.length > 0
+            }
+          >
+            <button className="btn btn-primary" onClick={() => updateStatus()}>
```
srkirkland, line 526: just do `<button className="btn btn-primary" onClick={updateStatus}>`

Hippo.Web/ClientApp/src/components/Order/OrderForm.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+  const [foundPI, setFoundPI] = useState(null);
+  const [localInstallmentType, setLocalInstallmentType] = useState(
+    methods.getValues("installmentType"),
+  );
+  const [limitedEditing, setLimitedEditing] = useState(false);
+
+  React.useEffect(() => {
+    const newStatus = orderProp.status;
+    methods.setValue("status", newStatus);
```
srkirkland, line 50: i think it'd be nicer to extract the `setValue` and `getValues` methods from `method` instead of using `methods.setValue` and then needing to useeffect depend on the entire method object.  Also, why call it methods, maybe at least formMethods but that doesn't really work either because formState is not a method.

Hippo.Web/Controllers/OrderController.cs
```diff
@@ -0,0 +1,866 @@
...
+            await _dbContext.SaveChangesAsync();
+
+            //To make sure the model has all the info needed to update the UI
+            var model = await _dbContext.Orders.Where(a => a.Cluster.Name == Cluster && a.Id == id)
+                .Include(a => a.MetaData).Include(a => a.Payments).Include(a => a.PrincipalInvestigator).ThenInclude(a => a.Owner)
+                .Include(a => a.History.Where(w => w.Type == History.HistoryTypes.Primary)).ThenInclude(a => a.ActedBy)
+                .Select(OrderDetailModel.Projection())
+                .SingleOrDefaultAsync();
+
+            return Ok(model);
```
laholstege, line 457: does all of this actually change? or is most of this already in the state for the UI and there's a smaller chunk of this we can update?

Hippo.Web/ClientApp/src/components/Product/Products.tsx
```diff
@@ -0,0 +1,462 @@
...
+                  lifeCycle: parseInt(e.target.value),
+                };
+                setEditProductModel(model);
+                setReturn(model);
+              }}
+            />
+          </div>
+        </>
+      ),
+      canConfirm:
```
srkirkland, line 223: feels like using the form library would make this validation logic easier and keep it consistent w/ the order form

Hippo.Web/Controllers/OrderController.cs
```diff
@@ -0,0 +1,866 @@
...
+
+        [HttpGet]
+        public async Task<IActionResult> Get(int id)
+        {
+            //TODO: When accounts has an IsActive flag, we will need to ignore the query filters.
+
+            var model = await _dbContext.Orders.Where(a => a.Cluster.Name == Cluster && a.Id == id)
+                .Include(a => a.MetaData).Include(a => a.Payments).Include(a => a.PrincipalInvestigator).ThenInclude(a => a.Owner)
+                .Include(a => a.History.Where(w => w.Type == History.HistoryTypes.Primary)).ThenInclude(a => a.ActedBy)
+                .Select(OrderDetailModel.Projection())
```
srkirkland, line 93: Include does nothing and is not necessary if you are doing a projection.  remove all includes.  Also, it's not a big deal but I don't think getting an order should by default include all history.  Seems like that should be a separate call, no?

Hippo.Web/Controllers/OrderController.cs
```diff
@@ -0,0 +1,866 @@
...
+            {
+                return Ok(new Account());
+            }
+
+            return Ok(model);
+
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> Save([FromBody] OrderPostModel model) //TODO: Might need to change this to a post model....
```
srkirkland, line 139: looks like you did

Hippo.Web/Controllers/OrderController.cs
```diff
@@ -0,0 +1,866 @@
...
+                await _dbContext.SaveChangesAsync(); 
+            }
+            catch (Exception ex)
+            {
+                return BadRequest(ex.Message);
+            }
+
+            //To make sure the model has all the info needed to update the UI
+            var rtmodel = await _dbContext.Orders.Where(a => a.Cluster.Name == Cluster && a.Id == model.Id)
+                .Include(a => a.MetaData).Include(a => a.Payments).Include(a => a.PrincipalInvestigator).ThenInclude(a => a.Owner)
```
srkirkland, line 264: remove includes

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
laholstege: i usually just put this in like a `Order.ts` file or wherever the model is

Hippo.Web/ClientApp/src/components/Order/Details.tsx
```diff
@@ -0,0 +1,685 @@
...
+
+  const paymentColumnHelper = createColumnHelper<PaymentModel>();
+
+  const paymentColumns = [
+    paymentColumnHelper.accessor("amount", {
+      header: "Amount",
+      id: "amount",
+      cell: (value) => (
+        <span>
+          <i className="fas fa-dollar-sign" />{" "}
```
laholstege, line 145:
```suggestion
          <FontAwesomeIcon icon={faDollarSign} />{" "}
```

Hippo.Web/ClientApp/src/components/Product/Products.tsx
```diff
@@ -0,0 +1,462 @@
...
+        setProducts(data);
+      } else {
+        alert("Error fetching products");
+      }
+    };
+
+    fetchProducts();
+  }, [cluster]);
+
+  const [getEditConfirmation] = useConfirmationDialog<ProductModel>(
```
laholstege, line 53: i think these should be <FormField>s inside a <Form>. little odd because the submit button in the modal will be outside of the form, but you can use a ref to submit

jSylvestre: I based these off code elsewhere in hippo. Seems to work as expected.

Hippo.Web/ClientApp/src/components/Order/Details.tsx
laholstege: but you'll have to import it too

Hippo.Web/ClientApp/src/components/Product/Products.tsx
laholstege: yep, i just mean for a consistent pattern. this is using our regular method of storing a local state and editing it `onChange`, but other places we're using `react-hook-form` to handle that. i can do that when i refactor, since seems to me this is also a form, just inside a modal. @srkirkland what do you think?

Hippo.Web/ClientApp/src/components/Order/BillingsFields.tsx
laholstege: @jSylvestre i think he just means `  const [notification, setNotification] = usePromiseNotification();` instead of a string

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
```diff
@@ -0,0 +1,175 @@
...
+  };
+
+  if (isClusterAdmin === null) {
+    return <div>Loading...</div>;
+  }
+
+  if (!order) {
+    return <div>Loading... {productId}</div>;
+  }
+
```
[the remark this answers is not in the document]
jSylvestre: I'd rather bounce them back to the products page if they fiddle with the URL

Hippo.Web/Controllers/OrderController.cs
jSylvestre: I had problems when I didn't do this.

Hippo.Web/ClientApp/src/components/Order/Details.tsx
jSylvestre: Maybe? the parseFloat below seems to have problems if I do.

Hippo.Web/ClientApp/src/components/Order/OrderForm.tsx
```diff
@@ -0,0 +1,343 @@
...
+    const newInstallmentDate = orderProp.installmentDate;
+    const newExpirationDate = orderProp.expirationDate;
+    methods.setValue("installmentDate", newInstallmentDate);
+    methods.setValue("expirationDate", newExpirationDate);
+  }, [
+    methods,
+    orderProp.expirationDate,
+    orderProp.installmentDate,
+    orderProp.status,
+  ]);
```
laholstege, line 72:
```suggestion
  const {
    handleSubmit,
    setError,
    setValue,
    getValues,
    formState: { isDirty, isSubmitting },
  } = methods;

  const [foundPI, setFoundPI] = useState(null);
  const [localInstallmentType, setLocalInstallmentType] = useState(
    getValues("installmentType"),
  );
  const [limitedEditing, setLimitedEditing] = useState(false);

  React.useEffect(() => {
    const newStatus = orderProp.status;
    setValue("status", newStatus);

    if (
      newStatus === "Active" ||
      newStatus === "Rejected" ||
      newStatus === "Cancelled" ||
      newStatus === "Completed"
    ) {
      setLimitedEditing(true);
    } else {
      setLimitedEditing(false);
    }

    const newInstallmentDate = orderProp.installmentDate;
    const newExpirationDate = orderProp.expirationDate;
    setValue("installmentDate", newInstallmentDate);
    setValue("expirationDate", newExpirationDate);
  }, [
    setValue,
    orderProp.expirationDate,
    orderProp.installmentDate,
    orderProp.status,
  ]);
```

the main difference here is that you have `setValue` instead of `methods` in your effect dependency array. this means that instead of the effect triggering every time the `methods` _object_ changes, you're just targeting the `setValue` property. potentially causing a lot less re-runs of the effect. normally, having an entire object in your dependency array will trigger the effect on every re-render, because of the way javascript handles object equality. 

```
const MyComponent = () => {
  const [state, setState] = React.useState({ key: 'value' });

  React.useEffect(() => {
    console.log('Effect ran');
  }, [state]);

  return (
    <div>
      <button onClick={() => setState({ key: 'value' })}>Update State</button>
    </div>
  );
};
```

the effect here will run every time the button is clicked, even though the value hasn't actually changed, because js using reference equality for objects, arrays, and functions. like, 

```
const obj1 = { a: 1 };
const obj2 = { a: 1 };
const obj3 = obj1;

console.log(obj1 === obj2); // false
console.log(obj1 === obj3); // true
```

so the object reference can change quite a lot on re-renders, even if the actual contents are the same. this is where `useMemo()` is useful, to say "this is the same object unless the dependencies change", and `useCallback()` similarly for functions.

so instead:
```
const MyComponent = () => {
  const [state, setState] = React.useState({ key: 'value' });

  React.useEffect(() => {
    console.log('Effect ran');
  }, [state.key]);

  return (
    <div>
      <button onClick={() => setState({ key: 'value' })}>Update State</button>
    </div>
  );
};
```
here, targeting the `key` only, the effect will only run when the _string_ 'value' changes, since in js `'value' === 'value'` returns true where `obj1 = { key: 'value'} === obj2 = {key: 'value'}` returns false. 

tl;dr it's safer to be as explicit in your deps as possible. possibly `react-hook-form` has thought of this and the `methods` object is memoized, but ya never know. also, javascript is weird

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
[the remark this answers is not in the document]
laholstege: you can still do that (and the effect will redirect them). i just like to handle potential error states both when fetching and when rendering. in your rendering logic right now, the case where `isClusterAdmin === false` isn't handled, so there's not fallback for what is actually displayed. it's not a big deal, especially in this case since i doubt we'll ever change the redirect, but it's a good habit to get into: both "how do i handle errors inside a function" and "what is rendered when there are errors". prob would only happen if something super weird happens, like the effect fails. but debugging stuff like that is a pain and having something specific rendered in that case is nice

Hippo.Web/ClientApp/src/components/Order/OrderForm.tsx
laholstege:
```jsx
  const {
    handleSubmit,
    setError,
    setValue,
    getValues,
    formState: { isDirty, isSubmitting },
  } = methods;

  const [foundPI, setFoundPI] = useState(null);
  const [localInstallmentType, setLocalInstallmentType] = useState(
    getValues("installmentType"),
  );
  const [limitedEditing, setLimitedEditing] = useState(false);

  React.useEffect(() => {
    const newStatus = orderProp.status;
    setValue("status", newStatus);

    if (
      newStatus === "Active" ||
      newStatus === "Rejected" ||
      newStatus === "Cancelled" ||
      newStatus === "Completed"
    ) {
      setLimitedEditing(true);
    } else {
      setLimitedEditing(false);
    }

    const newInstallmentDate = orderProp.installmentDate;
    const newExpirationDate = orderProp.expirationDate;
    setValue("installmentDate", newInstallmentDate);
    setValue("expirationDate", newExpirationDate);
  }, [
    setValue,
    orderProp.expirationDate,
    orderProp.installmentDate,
    orderProp.status,
  ]);
```

the main difference here is that you have `setValue` instead of `methods` in your effect dependency array. this means that instead of the effect triggering every time the `methods` _object_ changes, you're just targeting the `setValue` property. potentially causing a lot less re-runs of the effect. normally, having an entire object in your dependency array will trigger the effect on every re-render, because of the way javascript handles object equality. 

```jsx
const MyComponent = () => {
  const [state, setState] = React.useState({ key: 'value' });

  React.useEffect(() => {
    console.log('Effect ran');
  }, [state]);

  return (
    <div>
      <button onClick={() => setState({ key: 'value' })}>Update State</button>
    </div>
  );
};
```

the effect here will run every time the button is clicked, even though the value hasn't actually changed, because js using reference equality for objects, arrays, and functions. like, 

```js
const obj1 = { a: 1 };
const obj2 = { a: 1 };
const obj3 = obj1;

console.log(obj1 === obj2); // false
console.log(obj1 === obj3); // true
```

so the object reference can change quite a lot on re-renders, even if the actual contents are the same. this is where `useMemo()` is useful, to say "this is the same object unless the dependencies change", and `useCallback()` similarly for functions.

so instead:
```jsx
const MyComponent = () => {
  const [state, setState] = React.useState({ key: 'value' });

  React.useEffect(() => {
    console.log('Effect ran');
  }, [state.key]);

  return (
    <div>
      <button onClick={() => setState({ key: 'value' })}>Update State</button>
    </div>
  );
};
```
here, targeting the `key` only, the effect will only run when the _string_ 'value' changes, since in js `'value' === 'value'` returns true where `obj1 = { key: 'value'} === obj2 = {key: 'value'}` returns false. 

tl;dr it's safer to be as explicit in your deps as possible. possibly `react-hook-form` has thought of this and the `methods` object is memoized, but ya never know. also, javascript is weird

Hippo.Web/ClientApp/src/components/Order/BillingsFields.tsx
jSylvestre: I don't see how to get that to work. I only want to show a message for a short time if I've changed the natural account. Not every time validate the chart string is called.

Hippo.Web/ClientApp/src/components/Order/Details.tsx
jSylvestre: Can't remember for sure, but I think I was having trouble getting dynamic values for the status to work, and I just wanted one place where the status would get changed.

Hippo.Web/Controllers/OrderController.cs
jSylvestre: I think it is only getting the primary history when I do it like this and not all of it?
Anyway, will take a look. I see to remember having a problem...

jSylvestre: Maybe I did it without projection first... Will fix other places too

Hippo.Core/Domain/History.cs
```diff
@@ -75,5 +84,47 @@ public class Actions
...
+            public const string Processing = "Processing";
+            public const string Cancelled = "Cancelled";
+            public const string Active = "Active";
+            public const string Rejected = "Rejected";
+            public const string Completed = "Completed";
+            public const string AdhocPayment = "Adhoc Payment";
+            public const string ChartStringUpdated = "Chart String Updated";
+            public const string PaymentFailed = "Payment Failed";
+
+            public static List<string> TypeList = new List<string>
```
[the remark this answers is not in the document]
jSylvestre: Will change, but... I was following the pattern you had above :)

Hippo.Web/ClientApp/src/components/Order/Details.tsx
laholstege: #391

Hippo.Web/ClientApp/src/components/Order/Details.tsx
laholstege: #392

Hippo.Web/ClientApp/src/components/Order/Details.tsx
laholstege: #393

Hippo.Web/ClientApp/src/components/Order/OrderForm.tsx
laholstege: #394

Hippo.Web/ClientApp/src/components/Product/Products.tsx
laholstege: #395

Hippo.Web/ClientApp/src/components/Product/Products.tsx
laholstege: #395

Hippo.Web/ClientApp/src/components/Order/CreateOrder.tsx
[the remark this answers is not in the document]
laholstege: #396

Hippo.Web/ClientApp/src/components/Order/Details.tsx
laholstege: #395

PR #329, opened 2024-05-15
Jcs/flag for history display

jSylvestre:


Hippo.Web/Controllers/OrderController.cs
```diff
@@ -110,18 +110,9 @@ public async Task<IActionResult> CreateOrder([FromBody] Order model)
...
-            //    Action = History.OrderActions.Created,
-            //    Details = $"Order created by {currentUser.Email}" //Dump in json of the order? //Write if it is an adhoc order?
-            //};
-            //_dbContext.Histories.Add(history); //Add it here or call history service?
+
             await _historyService.OrderCreated(order, currentUser);
+            await _historyService.OrderSnapshot(order, currentUser, History.OrderActions.Created);
 
             _dbContext.Orders.Add(order);
             await _dbContext.SaveChangesAsync();
```
sprucely, line 118: Maybe do `await _dbContext.Orders.AddAsync(order);` before the two calls to `_historyService` in order to capture the order id in the Details json?

sprucely (changes requested):

Hippo.Web/Controllers/OrderController.cs
jSylvestre: I think I'd have to save it twice and I wanted to avoid that.
The history writes the correct value in the OrderId.

sprucely: Oh, I see. `await _historyService.OrderCreated(order, currentUser);` ends up updating the order id. In that case the call to `_dbContext.Orders.Add(order);` may be redundant. But I don't think it hurts to leave it there.

sprucely (approved):

PR #331, opened 2024-05-17
Do the validation when I get the order

jSylvestre:


srkirkland (changes requested):
`temp`, `temp2` and `BillingTemp` is too much temp for me

PR #334, opened 2024-05-20
Data serialization refactor

sprucely:
Some of the json Data columns were being passed to the client as strings because only the client is concerned with their contents, and because the data makes extensive use of unions, which aren't easily represented with C# types. But deserializing the Data properties client-side was ugly and error-prone. Making the Data type `JsonElement` seemed like a nice compromise, they're still loosely typed server-side, but no need to parse them client-side.

I temporarily generated a new migration to verify that the ValueConverter didn't change anything in the schema or the model snapshot. Also verified that it didn't impact the bulk operations performed by the `AccountSyncService`.


jSylvestre (dismissed):
Not sure that I understand all that...

srkirkland (approved):

PR #336, opened 2024-05-22
Jcs/new order fields20240522

jSylvestre:


Hippo.Core/Domain/Order.cs
```diff
@@ -34,6 +37,11 @@ public class Order
         [Range(1, int.MaxValue)]
         public int Installments { get; set; }
 
+        [Required]
+        [MaxLength(10)]
+        public string InstallmentType { get; set; } = "Monthly"; //Monthly, Yearly
```
srkirkland, line 42: use installmentTypes.Monthly here.  might have to change to static string but i think it should work

jSylvestre: @srkirkland Worked on the Product, guessing the Order would be fine too.

PR #338, opened 2024-05-23
Rh/order form typing / start of FieldArray

laholstege:


Hippo.Web/ClientApp/src/Shared/Form/FormFieldArray.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  console.log(arrayName);
+  return (
+    <>
+      {fields.map((field, index) => {
+        console.log("field", field);
+        console.log(`[arrayName][${index}]`, [arrayName]);
+        return (
+          <>
+            <FormField
```
jSylvestre, line 30: You would also need to show the value field
And possibly a button to add an empty array element and Delete?

jSylvestre (approved):

Hippo.Web/ClientApp/src/components/Order/MetaDataFields.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <Row key={field.id}>
+            <Col>
+              <FormField
+                key={`name-${field.id}`}
+                register={register}
+                label="Name"
+                error={errors.metaData?.[index]?.name}
+                name={`metaData.${index}.name`}
+                required={true}
+                autoComplete="nope"
```
srkirkland, line 48: wait you can just say `nope`?   like setting it to off but cooler?

Hippo.Web/ClientApp/src/components/Order/OrderForm.tsx
```diff
@@ -18,148 +19,130 @@ const OrderForm: React.FC<OrderFormProps> = ({
   readOnly,
   onSubmit,
 }) => {
+  const methods = useForm<OrderModel>({ defaultValues: orderProp });
```
srkirkland, line 22: doesn't really matter but why make this a two step assignment then destructure operation?

srkirkland (commented):
👍

Hippo.Web/ClientApp/src/components/Order/MetaDataFields.tsx
laholstege: nope
(this also didn't work)

Hippo.Web/ClientApp/src/components/Order/OrderForm.tsx
laholstege: @srkirkland because i want to spread all the methods as props to `<FormProvider>` but also select out those properties to use elsewhere. i could've done `methods.submitForm()` but destructuring 4 eva

PR #345, opened 2024-05-30
Upgrade react-table and add csv export

### LINKED ISSUE #296: Add a way for a PI to export the list of associated accounts to a spreadsheet (CSV?)
PI noted that being able to export the list of accounts in their group to a spreadsheet would be useful.

sprucely:
Upgrading wasn't absolutely necessary, but thought this would be a good time.

All data tables now have an `Export CSV` link and an optional `Export Filtered CSV` link that shows when a filter is applied. 
image
Also note the filter placeholders indicate the number of unique values in that column.

Complex/custom rendered cells can have the default export extraction overridden via `exportFn` in the columnDef's meta property:
```
      meta: {
        exportFn: (account) =>
          account.memberOfGroups.map((g) => g.displayName).join(", "),
      },
```
Likewise, the default filter control (`text`) can be changed to `range` or `select` via the `filterVariant` property. ()
```
      meta: {
        filterVariant: "select",
      },
```

closes #296


Hippo.Web/ClientApp/src/Shared/DebouncedInput.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+
+export const DebouncedInput = ({
```
srkirkland, line 3: this is interesting -- I can't tell from looking at it if there will be any issues w/ onChange not being useCallback() or other weird rendering side effects.  I've always just used the `useDebounce()` hook for this purpose (https://github.com/uidotdev/usehooks/blob/main/index.js#L239) which is simple to reason over.  If it works though I think it's a pretty nice solution.

srkirkland (approved):
code looks good, only thing I don't like is the showing of number of unique values in the filter header.  I think it'll just be confusing and isn't very helpful.  I like how you built the export functionality into the shared table itself -- is there any way to turn it off if you don't want a table to be exportable?

PR #347, opened 2024-06-06
Jcs/link to accounts instead

jSylvestre:


Hippo.Core/Domain/Order.cs
```diff
@@ -110,7 +110,11 @@ internal static void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<Billing>().HasOne(o => o.Order).WithMany(o => o.Billings).HasForeignKey(o => o.OrderId).OnDelete(DeleteBehavior.Restrict);
             modelBuilder.Entity<OrderMetaData>().HasOne(o => o.Order).WithMany(o => o.MetaData).HasForeignKey(o => o.OrderId).OnDelete(DeleteBehavior.Restrict);
             modelBuilder.Entity<Payment>().HasOne(o => o.Order).WithMany(o => o.Payments).HasForeignKey(o => o.OrderId).OnDelete(DeleteBehavior.Restrict);
-
+            modelBuilder.Entity<Order>()
+                .HasOne(o => o.PrincipalInvestigator)
+                .WithMany(u => u.Orders)
+                .HasForeignKey(o => o.PrincipalInvestigatorId)
+                .OnDelete(DeleteBehavior.Restrict);
```
sprucely, line 117: I'm confused. You're declaring `DeleteBehavior.Restrict` here, but the migrations are showing `onDelete: ReferentialAction.Cascade`. That doesn't seem right.

sprucely (changes requested):
Not sure we should have any cascading deletes. Or is that just temporary until we add IsActive to accounts?

PR #367, opened 2024-07-11
Refactor account sync (merge all clusters at once)

sprucely:
I verified that it is correctly removing group memberships.

I wasn't sure about using a tuple `(a.ClusterId, a.Kerberos)` as a dictionary key, but it seemed cleaner than defining a wrapper type and manually implementing all the hash/equality stuff to be able to use it as a key.


laholstege (approved):

jSylvestre (approved):
Didn't test.

PR #373, opened 2024-07-12
Show/Hide order info per cluster

jSylvestre:


Hippo.Core/Services/UserService.cs
```diff
@@ -214,7 +214,7 @@ public async Task<User> GetUser(Claim[] userClaims)
 
         public async Task<string> GetAvailableClustersJsonAsync()
         {
-            var clusters = await _dbContext.Clusters
+            var clusters = await _dbContext.Clusters.Include(a => a.FinancialDetail)
```
sprucely, line 217: I don't think `Include(...)` is necessary when you do a projection.

sprucely (approved):
Seems okay to me. I don't think the call to Include does anything in this situation.

PR #378, opened 2024-07-17
Expand qos references during account sync

### LINKED ISSUE #365: Expand QOS references during account sync
Image
`gpum-users-gpum-qos` is not informative for PIs. It is a reference to `group::gpum-users::slurm::partitions::gpum` within the yaml file that data came from. Find a way to expand that reference in place at the time account/group data is synced.

The format of the reference is `[GROUP]-[PARTITION]-qos`. GROUP can have additional hyphens.

sprucely:
Before:
image

After:
image

Closes #365

### DIFF
--- a/Hippo.Core/Services/AccountSyncService.cs
+++ b/Hippo.Core/Services/AccountSyncService.cs
@@ -1,5 +1,8 @@
 
+using System.Collections.Concurrent;
 using System.ComponentModel.DataAnnotations;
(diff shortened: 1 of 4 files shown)


jSylvestre (approved):

srkirkland (approved):
a lot of trickiness, looks like you handled possible cases of bad input data pretty well which is good

PR #382, opened 2024-07-18
New job for order processing

jSylvestre:


sprucely (approved):
Looks good to me. I know it's not going into main yet. But once it does, be aware that hippo-test is configured with `WEBJOBS_DISABLE_SCHEDULE = 1`, so no jobs are running automatically.

Hippo.Jobs.OrderProcess/run.cmd
```diff
@@ -0,0 +1,3 @@
+@echo off
+
+dotnet Hippo.Jobs.OderProcess.dll
```
srkirkland, line 3: typo

PR #384, opened 2024-07-19
Leave sshkey out of yaml when disabled/not relevant

### LINKED ISSUE #289: Leave sshkey out of yaml when disabled

sprucely:
Not sure if there's any way to really test this. It should be safe to rely on only the presence of `account.Key`, since that was thoroughly validated at the time an account was requested...

https://github.com/ucdavis/hippo/blob/05086736d396b85df548b9935ad0415031536ef0/Hippo.Web/Controllers/AccountController.cs#L108-L137

Closes #289

### DIFF
--- a/Hippo.Core/Services/AccountUpdateYamlService.cs
+++ b/Hippo.Core/Services/AccountUpdateYamlService.cs
@@ -47,23 +47,41 @@ public class AccountUpdateYamlService : IAccountUpdateService
         var group = queuedEventModel.Data.Groups.SingleOrDefault();
         var serializer = new Serializer();
         return serializer.Serialize(
-            new
-            {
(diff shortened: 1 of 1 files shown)


srkirkland (changes requested):
that's basically the same thing just repeated, I think it'd be cleaner to just add the isNullOrWhiteSpace on the key prop, like 
```
                account = new
                {
                    name = account.Name,
                    email = account.Email,
                    kerb = account.Kerberos,
                    iam = account.Iam,
                    mothra = account.Mothra,
                    key =  account.Key ?? ""
                },
                ```

sprucely:
@srkirkland That's not what they wanted. They asked for the key property to not be included in the yaml instead of leaving an undefined, null or empty string.

srkirkland (approved):

Issue #402, opened 2024-07-25
History


jSylvestre:
System was showing in history when admin approved order
Maybe because the react change I made or might be projection needs the user.

jSylvestre:
The history record has an acted by, so it is either my projection or react code changes

jSylvestre:
For the order projection, I filter the history, but unless I include the hostory.actedBy, it returns null and System is displayed.
Can either do an Include before the projection, or maybe have a separate call that populates the history table.

PR #407, opened 2024-07-26
Jcs/history and payment endpoints

jSylvestre:


jSylvestre:
When this is used, it should fix #402

PR #414, opened 2024-07-30
Add account tags

### LINKED ISSUE #297: Allow PIs to add tags to their users
PI noted they would like to be able to "add their own tags" to users, presumably as a way to keep track of user association within their group. Class account vs sub-lab account?

sprucely:
Eventually, we'll want to be able to edit, merge and delete existing tags. But I don't think that's a high priority at this point.

image

image

image

closes #297

### DIFF
--- a/Hippo.Core/Domain/Tag.cs
+++ b/Hippo.Core/Domain/Tag.cs
@@ -1,0 +1,16 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace Hippo.Core.Domain;
+
+public class Tag
+{
+    public int Id { get; set; }
+    [MaxLength(50)]
+    [Required]
+    public string Name { get; set; }
+
+    public int ClusterId { get; set; }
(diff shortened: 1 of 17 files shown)


Hippo.Core/Domain/Tag.cs
```diff
@@ -0,0 +1,16 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace Hippo.Core.Domain;
+
+public class Tag
+{
+    public int Id { get; set; }
```
jSylvestre, line 7: Should this have the [Key] attribute?

jSylvestre (approved):
Didn't test.
Just had the one comment.

Hippo.Core/Domain/Tag.cs
sprucely: Not really necessary, since it's covered by convention.

PR #417, opened 2024-08-01
Payment Notification

jSylvestre:


srkirkland (commented):
i'll leave this for @sprucely to look at when he fixes the email stuff.  Though if you want to merge anyway and fix later I think this looks fine.

sprucely (approved):
Yeah, having the extra email models was one of the issues I had with the dependency direction. But this is fine for now. I can clean it up later.

Issue #413, opened 2024-07-30
Order List filtering


jSylvestre:
The ID, and  CreatedOn columns don't seem to filter correctly

jSylvestre:
Created on works if you type yyyy-mm-dd

PR #418, opened 2024-08-01
Search the order id in the table

jSylvestre:
Issue #413


sprucely (approved):
Looks fine. I don't know if case sensitivity matters for ID. If so, maybe do something like...
`return row.original.id.toString().toLower().includes(filterValue.toLower());`

jSylvestre:
> Looks fine. I don't know if case sensitivity matters for ID. If so, maybe do something like... `return row.original.id.toString().toLower().includes(filterValue.toLower());`

Just an int, so no.
Could potentially have it do an exact match, but I figured this would be a little more flexible. No idea why it needed this though

PR #419, opened 2024-08-01
status bar

laholstege:
tried to consolidate status logic into one place (`status.ts`) 

styling for the form is not done, but this does the status bar. indicates on hover when an action will change the status 

image

https://github.com/user-attachments/assets/0bd99434-0096-4ba2-9b0e-27c4786881b6

https://github.com/user-attachments/assets/2836400f-194c-4729-8f17-ad9ee3d196f1

### DIFF
--- a/Hippo.Web/ClientApp/src/components/Account/AccountStatus.tsx
+++ b/Hippo.Web/ClientApp/src/components/Account/AccountStatus.tsx
@@ -2,6 +2,7 @@ import { useContext } from "react";
 import AppContext from "../../Shared/AppContext";
 import { useParams } from "react-router-dom";
 import { SplitCamelCase } from "../../util/StringHelpers";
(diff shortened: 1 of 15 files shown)


Hippo.Web/ClientApp/src/components/Account/AccountStatus.tsx
```diff
@@ -26,14 +27,14 @@ export const AccountStatus = () => {
               {SplitCamelCase(request.status)}
             </span>
             .
-            {request.status === "PendingApproval" && (
+            {request.status === RequestStatus.PendingApproval && (
```
srkirkland, line 30: 👍

srkirkland (approved):
i like it, nice.  especially getting the statuses defined is great to have

jSylvestre (approved):

PR #449, opened 2024-08-19
Is Recurring DB change

jSylvestre:


sprucely (approved):
Looks good. Nothing jumping out at me.

PR #457, opened 2024-08-27
Cyd/checkboxes

cydoval:
moving checkboxes to their own row


jSylvestre (approved):
Style isn't applied to the financial details page, but looks better on the product and order pages.

PR #459, opened 2024-08-27
Allow UserService to handle empty claims collection

sprucely:


jSylvestre (approved):
Seems safe enough, but wouldn't it be better not to call the GetUser at all if an auth token is provided?

sprucely:
> Seems safe enough, but wouldn't it be better not to call the GetUser at all if an auth token is provided?

It's the `HistoryService.AddHistory(...)` that calls `_userService.GetCurrentUser()` . It has no knowledge of auth token.

PR #461, opened 2024-08-29
When adding an admin, add an cluster account if not there

jSylvestre:


Hippo.Web/Controllers/OrderController.cs
```diff
@@ -72,10 +72,10 @@ public async Task<IActionResult> AdminOrders()
...
             }
             var currentUserAccount = await _dbContext.Accounts.SingleOrDefaultAsync(a => a.Cluster.Name == Cluster && a.OwnerId == currentUser.Id);
-            if (currentUserAccount == null)
-            {
-                return Ok(new OrderListModel[0]);
-            }
+            //if (currentUserAccount == null)
+            //{
+            //    return Ok(new OrderListModel[0]);
+            //}
```
sprucely, line 78: Should probably just remove this, since currentUserAccount is no longer referenced.

Hippo.Web/Controllers/OrderController.cs
```diff
@@ -289,7 +289,7 @@ public async Task<IActionResult> UpdateBilling([FromBody] OrderPostModel model)
 
             //TODO: Validation
             //Updating an existing order without changing the status.
-            var existingOrder = await _dbContext.Orders.Include(a => a.PrincipalInvestigator).Include(a => a.Cluster).Include(a => a.Billings).FirstAsync(a => a.Id == model.Id);
+            var existingOrder = await _dbContext.Orders.Include(a => a.PrincipalInvestigator.Owner).Include(a => a.Cluster).Include(a => a.Billings).FirstAsync(a => a.Id == model.Id);
```
sprucely, line 292: Perhaps rename the `a` variables in those `Include(...)` calls to `o`, since they're called on Order collections?

sprucely (approved):
Looks good. Just some nitpicks.

PR #463, opened 2024-08-30
Check for negative balance.

jSylvestre:
I had one in test, but it think the balance remaining on that one was wrong because I changed where the balance remaining was updated, so it got the old way plus the new way.


srkirkland (approved):
worth setting a history entry saying total balance was adjusted to 0?

PR #466, opened 2024-09-03
Fix UTC logic

jSylvestre:


jSylvestre:
Next Payment Date is:
2024-10-01 07:00:00.0000000

Previous payment date was:
2024-09-01 06:00:50.6180677

srkirkland (approved):

PR #469, opened 2024-09-05
Jcs/financial role

jSylvestre:


sprucely (approved):
Didn't test, but looks fine.

Hippo.Web/Controllers/AdminController.cs
```diff
@@ -127,6 +127,96 @@ public async Task<IActionResult> RemoveClusterAdmin(int id)
...
+        // get all users with cluster admin permissions
+        return Ok(await _dbContext.Users
+            .AsNoTracking()
+            .Where(u => u.Permissions.Any(p => p.Cluster.Name == Cluster && p.Role.Name == Role.Codes.FinancialAdmin))
+            .OrderBy(u => u.LastName).ThenBy(u => u.FirstName)
+            .ToArrayAsync());
+    }
+
+    [HttpPost]
+    public async Task<IActionResult> AddFinancialAdmin(string id)
```
srkirkland, line 143: seems very repetitive w/ add/remove cluster admin, almost copy/paste could probably be combined

jSylvestre: Yeah... I got lazy, it is a little different query...

srkirkland (commented):
definitely better.  maybe next time just something like `RemoveAdminRole?type=financial` so it can be extended in the future.  we can catch it next time though

PR #473, opened 2024-09-06
use Effect instead of state

laholstege:


jSylvestre (approved):
Didn't test. But looks better.

jSylvestre (changes requested):
Testing shows some problems.

jSylvestre:
PR #477 fixes the issue I've noticed originally. Not sure if that will cause conflicts with this PR, but I figure it can be fixed later.

laholstege:
if it aint broke

PR #488, opened 2024-10-04
Add Notify API

sprucely:
Moved `Hippo.Email` code into `Hippo.Core`
`/api/notify/raw` allows client to explicitly set text and optionally html body.
`/api/notify/styled` creates a styled email via `NotificationService.SimpleNotification`.

image

image


jSylvestre (approved):
Didn't test, but looked good to me.

PR #510, opened 2025-01-08
Jcs/record payment details

jSylvestre:


sprucely (approved):
Looks fine to me. But it might be good to copy some serialization settings, like `camelCase`, in case we end up wanting to do something with that data in the front end.

PR #522, opened 2025-01-15
Cyd/navigation menus

jSylvestre:


jSylvestre:
Group Admin:
image
Group Admin with Orders Active:
image
Cluster Admin with orders active:
image
image
image
Financial Admin with orders active:
image
image

sprucely (approved):

PR #525, opened 2025-01-15
Cluster admins should not see all

jSylvestre:


Hippo.Web/ClientApp/src/Shared/ShowFor.tsx
```diff
@@ -51,13 +51,23 @@ export const ShowFor = (props: Props) => {
 
   // cluster admins can access anything in their cluster
   if (
+    roles.some((r) => r === "ClusterAdmin") &&
     context.user.permissions.some(
       (p) => p.cluster === cluster && p.role === "ClusterAdmin",
     )
   ) {
     return <>{children}</>;
   }
```
sprucely, line 60: I don't think the comment, about cluster admins accessing anything, is still applicable with this change.

sprucely (approved):
Didn't test, but I think that looks right.

PR #529, opened 2025-01-21
Jcs/notification job20250116

jSylvestre:


cbookman:
<_Order_> will end its life span on <date>.

To avoid any downtime, you may want to contact the cluster admin(s) to purchase the new equipment, or you can use the Order Replacement button below.

Hippo.Core/Services/NotificationService.cs
```diff
@@ -348,6 +350,44 @@ await _emailService.SendEmail(new EmailModel
             }
         }
 
+        public async Task<bool> OrderNotificationTwoButton(SimpleNotificationModel simpleNotificationModel, Order order, string[] emails, string[] ccEmails = null)
+        {
+            try
+            {
+                var message = simpleNotificationModel.Paragraphs.FirstOrDefault();
```
sprucely, line 357: Why only the first paragraph for the text body? I would have joined them into a single string joined by `[CRLF][CRLF]`. Is it because order notifications won't ever have more than one paragraph?

sprucely (approved):
Code looks fine. Just curious about the text body truncation.

PR #536, opened 2025-02-06
Add system and api endpoints to trigger puppet account sync

### LINKED ISSUE #500: GitHub Sync API Endpoint
There are sometimes cases where we want to make sure sponsorships and group memberships get synced from the Cheeto DB to HiPPO ASAP. One possibility is to add another endpoint to the API where we could post a request to kick off a sync. Obviously this would need to be a use-with-care sort of thing, or would need some rate limiting.

sprucely:
Closes #500


jSylvestre (approved):
Didn't test

PR #540, opened 2025-02-12
Add "Request Create Group" feature

### LINKED ISSUE #499: "Request Lab Group" Button and Event Type
The current workflow for PIs to request a group for themselves is unintuitive and requires some gross implicit logic on the backend. Instead of having them apply to a special "Sponsors" group to denote that they want a group of their own, it would be great if HiPPO just had a button for "Request Group Creation" or "Request Lab Group," with a corresponding NewGroup event type for the queue. I imagine that this would still use the `sponsors` group internally to track who gets the approval emails and whatnot, but it would make things much more clear for the users, and a dedicated event type would remove some annoying implicit magic in cheeto on our end.

sprucely:
Touched a lot of parts, but I don't think I went too crazy with changes. I had to update Serilog dependencies and add a package to fix a bug in how the Exceptions sink handles SqlServerExceptions. In the process, I discovered the Elasticsearch sink is deprecated. We'll have to replace that in a future update, probably whenever we upgrade to .Net8.

Closes #499

### DIFF
--- a/Hippo.Core/Domain/QueuedEvent.cs
+++ b/Hippo.Core/Domain/QueuedEvent.cs
@@ -38,10 +38,12 @@ public class QueuedEvent
         public const string CreateAccount = "CreateAccount";
         public const string UpdateSshKey = "UpdateSshKey";
         public const string AddAccountToGroup = "AddAccountToGroup";
+        public const string CreateGroup = "CreateGroup";
 
         public const string RegexPattern = CreateAccount
             + "|" + UpdateSshKey
-            + "|" + AddAccountToGroup;
+            + "|" + AddAccountToGroup
+            + "|" + CreateGroup;
     }
 
     public static class Statuses
--- a/Hippo.Core/Services/AccountUpdateService.cs
+++ b/Hippo.Core/Services/AccountUpdateService.cs
@@ -9,6 +9,7 @@ using Hippo.Core.Models;
 using Microsoft.EntityFrameworkCore;
 using Serilog;
 using YamlDotNet.Serialization;
+using Hippo.Core.Extensions;
 
 namespace Hippo.Core.Services
 {
@@ -53,6 +54,7 @@ namespace Hippo.Core.Services
                     QueuedEvent.Actions.UpdateSshKey => await CompleteUpdateSshKey(queuedEvent.Data),
                     QueuedEvent.Actions.CreateAccount => await CompleteCreateAccount(queuedEvent.Data),
                     QueuedEvent.Actions.AddAccountToGroup => await CompleteAddAccountToGroup(queuedEvent.Data),
+                    QueuedEvent.Actions.CreateGroup => await CompleteCreateGroup(queuedEvent.Data),
                     _ => Result.Error("Unknown action: {Action}", queuedEvent.Action)
                 };
             }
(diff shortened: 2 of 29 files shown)


Hippo.Core/Domain/QueuedEvent.cs
```diff
@@ -38,10 +38,12 @@ public static class Actions
         public const string CreateAccount = "CreateAccount";
         public const string UpdateSshKey = "UpdateSshKey";
         public const string AddAccountToGroup = "AddAccountToGroup";
+        public const string CreateGroup = "CreateGroup";
 
         public const string RegexPattern = CreateAccount
             + "|" + UpdateSshKey
-            + "|" + AddAccountToGroup;
+            + "|" + AddAccountToGroup
+            + "|" + CreateGroup;
```
srkirkland, line 46: Getting close to the point where a string.join() with a pipe delimiter would be cleaner.

Hippo.Core/Services/AccountUpdateService.cs
```diff
@@ -177,6 +179,44 @@ private async Task<Result> CompleteCreateAccount(QueuedEventDataModel data)
...
+            var newGroup = new Group
+            {
+                Name = groupModel.Name,
+                DisplayName = data.Metadata?["DisplayName"] ?? groupModel.Name,
+                AdminAccounts = new () { account },
+                ClusterId = account.ClusterId
+            };
+            
+
+            await _dbContext.Groups.AddAsync(newGroup);
```
srkirkland, line 216: Just checking to make sure you don't need to save changes here. Might be doing it elsewhere

Hippo.Core/Services/NotificationService.cs
```diff
@@ -14,13 +14,16 @@
 using Razor.Templating.Core;
 using Mjml.Net;
 using Hippo.Core.Models;
+using StrawberryShake.Json;
```
srkirkland, line 17: What's this for? Isn't it a graphQL thing?

Hippo.Web/Controllers/GroupController.cs
```diff
@@ -170,6 +170,89 @@ public async Task<IActionResult> RequestAccess([FromBody] AddToGroupModel addToG
...
+        }
+
+#pragma warning disable CS8600 // Converting null literal or possible null value to non-nullable type.
+        var alreadyRequested = await _dbContext.Requests.AnyAsync(r =>
+            r.Cluster.Name == Cluster
+            && r.Requester.Iam == currentUser.Iam
+            && r.Action == HippoRequest.Actions.CreateGroup
+            && HippoRequest.Statuses.Pending.Contains(r.Status)
+            // this ugly cast is an unfortunate side effect of automapping nvarchar to a JsonElement property
+            && CustomFunctions.JsonValue((string)(object)r.Data, "$.name") == createGroupModel.Name
```
srkirkland, line 215: Either genius or madness

Hippo.Web/Controllers/RequestController.cs
```diff
@@ -227,23 +272,42 @@ public async Task<ActionResult> Reject(int id, [FromBody] RequestRejectionModel
...
         {
-            Log.Error("Error creating Account Decision email");
+            case HippoRequest.Actions.AddAccountToGroup:
+            case HippoRequest.Actions.CreateAccount:
+                var group = await _dbContext.Groups.SingleAsync(g => g.ClusterId == request.ClusterId && g.Name == request.Group);
+                isGroupAdmin = await _dbContext.GroupAdminAccount.AnyAsync(ga =>
+                    ga.GroupId == group.Id
+                    && ga.Group.AdminAccounts.Any(aa => aa.OwnerId == currentUser.Id));
+
+                if (!isClusterOrSystemAdmin && !isGroupAdmin)
```
srkirkland, line 286: Is this supposed to be an and?

srkirkland (approved):

Hippo.Core/Services/AccountUpdateService.cs
sprucely: It's done by the caller. I'll add a comment about it...

Hippo.Core/Services/NotificationService.cs
sprucely: Not sure how that got in there. Cleaning it up...

jSylvestre: We do ChartString validation, but I don't think that is needed there.

Hippo.Core/Domain/QueuedEvent.cs
sprucely: Can't do that. It needs to be a `const` expression in order to be referenced from attributes like `[StringOptions(QueuedEvent.Actions.RegexPattern)]`

PR #551, opened 2025-03-10
Jcs/problem orders

### LINKED ISSUE #541: Create a problem orders view
Check if Active orders have next billing in the past and no payment record
Check if active orders have chart strings with errors or warnings

jSylvestre:
close #541

image


sprucely (approved):
Didn't test, but looks good.

PR #560, opened 2025-03-20
 Fix date sorting on some tables 

### LINKED ISSUE #538: "Updated on" sort field sorts numerically instead of by date.
Clicking the 'Updated On' in Active Accounts sorts items numerically instead of by actual date (ascending or descending.

sprucely:
Closes #538


srkirkland (approved):
put it into a helper

PR #564, opened 2025-04-09
Don't touch AcceptableUsePolicyAgreedOn during account sync

sprucely:
I tested the sync job locally.


jSylvestre (approved):
didn't test, but looks correct.

PR #565, opened 2025-04-10
Disable account delete and normalize account.data

### LINKED ISSUE #539: Account `UpdatedOn` appears to be updated at every puppet sync

sprucely:
I haven't actually been able to reproduce accounts being updated every day. What I think may be happening is a group with many accounts gets its QOS updated, resulting in every member account getting a corresponding update to its Json Data property, since QOS references are expanded during sync. So the json normalization is probably unnecessary, but it doesn't put much of a dent in performance.

Closes #539


jSylvestre (approved):
Didn't test, but looks good to me

PR #567, opened 2025-04-16
Swe/account soft deletes

### LINKED ISSUE #537: Ensure account sync won't break when accounts are deleted from puppet
I believe sync will break the moment an account is deleted from puppet, since it appears a number of foreign keys to Hippo's Accounts table have been added since account sync was implemented.

sprucely:
Closes #537

### DIFF
--- a/Hippo.Core/Domain/Order.cs
+++ b/Hippo.Core/Domain/Order.cs
@@ -95,7 +95,7 @@ namespace Hippo.Core.Domain
         }
         internal static void OnModelCreating(ModelBuilder modelBuilder)
         {
-            modelBuilder.Entity<Order>().HasQueryFilter(o =>o.PrincipalInvestigator.IsActive && o.Cluster.IsActive);
+            modelBuilder.Entity<Order>().HasQueryFilter(o => o.PrincipalInvestigator.IsActive && o.Cluster.IsActive);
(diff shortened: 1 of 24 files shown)


Hippo.Core/Domain/Order.cs
```diff
@@ -95,7 +95,7 @@ public class Statuses
         }
         internal static void OnModelCreating(ModelBuilder modelBuilder)
         {
-            modelBuilder.Entity<Order>().HasQueryFilter(o => o.Cluster.IsActive);
+            modelBuilder.Entity<Order>().HasQueryFilter(o =>o.PrincipalInvestigator.IsActive && o.Cluster.IsActive);
```
jSylvestre, line 98: Not sure this is correct

srkirkland (approved):

Hippo.Core/Domain/OrderMetaData.cs
```diff
@@ -27,6 +27,7 @@ public class OrderMetaData
 
         internal static void OnModelCreating(ModelBuilder modelBuilder)
         {
+            modelBuilder.Entity<OrderMetaData>().HasQueryFilter(omd => omd.Order.Cluster.IsActive);
```
jSylvestre, line 30: Same here. 
Is a cluster going to be inactive once we start taking orders?

PR #571, opened 2025-05-30
Update feature flags

sprucely:
Structured it to be easily extendable in order to facilitate more frequent pushes to prod.


jSylvestre (approved):
Didn't test, but looks good to me.

PR #604, opened 2026-04-06
Let cluster admins admin

### LINKED ISSUE #601: Cluster Admin can approve Create Account requests, but not Create Group requests

sprucely:
Closes #601

### DIFF
--- a/Hippo.Web/ClientApp/src/Shared/usePermissions.ts
+++ b/Hippo.Web/ClientApp/src/Shared/usePermissions.ts
@@ -12,8 +12,8 @@ export const usePermissions = () => {
   );
 
   const canViewGroup = (groupName: string) => {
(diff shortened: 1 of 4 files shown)


Hippo.Web/ClientApp/src/Shared/usePermissions.ts
```diff
@@ -22,8 +22,8 @@ export const usePermissions = () => {
   };
 
   const canManageGroup = (groupName: string) => {
-    if (isSystemAdmin) return true;
     if (!clusterName) return false;
+    if (isSystemAdmin || isClusterAdmin) return true;
```
jSylvestre, line 26: Is there anything the system admin allows that we don't want a cluster admin to do?

sprucely: I don't think so.

PR #618, opened 2026-07-27
Fix user/account issues

sprucely:
## Summary

- Improves account ownership linking when users and synced accounts share Kerberos IDs.
- Avoids assigning ownership when Kerberos matches are ambiguous.
- Preserves existing account owners during sync instead of clearing them for duplicate Kerberos cases.
- Validates queued create-account events so IAM/Kerberos mismatches fail safely.
- Adds regression tests for ownership linking, sync ownership preservation, and queued event validation.

## Summary by CodeRabbit

* **New Features**
  * Automatically links accounts to matching users when ownership is unassigned, including during group and account update flows.
  * Improves ownership selection during account synchronization to handle ambiguous identity data while preserving existing ownership when appropriate.
* **Bug Fixes**
  * Avoids hard failures on non-unique Kerberos lookups; ambiguous matches are skipped and warnings are logged.
  * Prevents overwriting an account’s existing owner during linking.
* **Tests**
  * Adds/expands unit tests for ownership linking, synchronization ownership mapping, account update create flows, and user provisioning.
  * Introduces SQLite-backed test context support for more reliable service tests.

### DIFF
--- a/Hippo.Core/Services/AccountOwnershipService.cs
+++ b/Hippo.Core/Services/AccountOwnershipService.cs
@@ -1,6 +1,7 @@
 using Hippo.Core.Data;
 using Hippo.Core.Domain;
 using Microsoft.EntityFrameworkCore;
+using Serilog;
 
 namespace Hippo.Core.Services;
 
@@ -10,6 +11,15 @@ public static class AccountOwnershipService
     {
         if (string.IsNullOrWhiteSpace(user.Kerberos))
         {
+            return 0;
+        }
+
+        if (await HasMultipleUsersWithKerberos(dbContext, user.Kerberos))
+        {
+            Log.Warning(
+                "Multiple users found with Kerberos {Kerberos}. Skipping account ownership linking for user {UserId}.",
+                user.Kerberos,
+                user.Id);
             return 0;
         }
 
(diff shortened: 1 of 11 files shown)


Hippo.Core/Services/AccountOwnershipService.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        if (string.IsNullOrWhiteSpace(user.Kerberos))
+        {
+            return 0;
+        }
+
+        if (await HasConflictingUserWithKerberos(dbContext, user))
+        {
+            Log.Warning(
+                "Multiple users found with Kerberos {Kerberos}. Skipping account ownership linking for user {UserId}.",
```
jSylvestre: If this is a new user that would cause a dup, the log doesn't indicate that.
So maybe User's kerb {} would cause duplication. Skipping...

Hippo.Core/Services/AccountSyncService.cs
```diff
@@ -372,5 +391,52 @@ public static List<GroupMemberAccount> ApplyRevokedGroupMemberAccounts(
...
+        }
+
+        public static int? GetDesiredOwnerId(
+            int clusterId,
+            string kerberos,
+            IReadOnlyDictionary<string, int> uniqueUserIdsByKerberos,
+            IReadOnlySet<string> ambiguousKerberos,
+            IReadOnlyDictionary<(int ClusterId, string Kerberos), int?> existingOwnerIdsByAccountKey)
+        {
+            if (uniqueUserIdsByKerberos.TryGetValue(kerberos, out var userId))
```
jSylvestre, line 468: Codex comment:
[P1] Sync can overwrite an existing account owner. [AccountSyncService.cs (line 431)](C:\\GitProjects\\hippo\\Hippo.Core\\Services\\AccountSyncService.cs:431) returns the unique Kerberos match before considering the existing owner. Because OwnerId now participates in bulk-update comparisons, a sync can reassign an account from user A to user B. Preserve a non-null existing owner and only infer ownership for unowned accounts.

Hippo.Core/Services/IdentityService.cs
```diff
@@ -60,9 +61,14 @@ public async Task<User> GetByKerberos(string kerb)
             {
                 var iamIds = ucdKerbResult.ResponseData.Results.Select(a => a.IamId).Distinct().ToArray();
                 var userIDs = ucdKerbResult.ResponseData.Results.Select(a => a.UserId).Distinct().ToArray();
-                if (iamIds.Length != 1 && userIDs.Length != 1)
+                if (iamIds.Length != 1 || userIDs.Length != 1)
                 {
```
jSylvestre, line 65: What if either of these returns zero?

jSylvestre (approved):
Couple comments for you to review.
Didn't test, but looks ok to me.

Hippo.Core/Services/IdentityService.cs
sprucely: They won't, those arrays are projected from a non-empty result set

